PR #1, opened 2019-02-19
Scroll to newly selected component if it's out of view

Jessidhia:


src/devtools/views/Element.js
```diff
@@ -45,6 +52,13 @@ export default function ElementView({ index, style }: Props) {
   const showDollarR =
     isSelected && (type === ElementTypeClass || type === ElementTypeFunction);
 
+  const component = useRef(null);
+  useEffect(() => {
+    if (isSelected && component.current !== null) {
+      component.current.scrollIntoView();
```
Jessidhia: It sometimes looks nice with `{ behavior: 'smooth' }`, sometimes looks jarring 🤔

bvaughn:
[This behavior is meant to be managed by react-window's list component.](https://github.com/bvaughn/react-devtools-experimental/blob/a31415011d46d7be3302ad12777df04170adb382/src/devtools/views/Tree.js#L37-L43) Elements (rows) aren't meant to scroll themselves.

If you see a case that isn't working, repro steps (or a fix to the `Tree` component check) would be good though. 😄

Jessidhia:
Oh. I'll check that out tomorrow then 👍

FWIW I was trying to make it easier to use on very large trees; maybe it's a problem that only happens if it's deep enough to go off the right side.

bvaughn:
Ah, that makes sense. So it scrolls to the correct vertical place but it's not correct horizontally?

Interesting. Let's keep this one open for now then and think about it 👍

Jessidhia:
It's not quite the correct vertical place either. This might show it:

image

It scrolls but it's like 90% obscured by the horizontal scrollbar (that thin blue line is the selected element's indent color), and there is no horizontal scrolling at all.

bvaughn:
Thanks :)

PR #2, opened 2019-02-19
Fix encoding of Unicode keys greater than U+00FF

Jessidhia:
Before:
```js
utfDecodeString(utfEncodeString("脳みそ爆発🤯"))
// => "3]z>/"
```

After:

```js
utfDecodeString(utfEncodeString("脳みそ爆発🤯"))
// => "脳みそ爆発🤯"
```

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -51,7 +51,7 @@ export function getUID(): number {
   return ++uidCounter;
 }
 
-export function utfDecodeString(array: Uint8Array): string {
+export function utfDecodeString(array: Uint16Array): string {
   let string = '';
   const { length } = array;
   for (let i = 0; i < length; i++) {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -51,20 +51,15 @@ export function getUID(): number {
...
-export function utfEncodeString(string: string): Uint8Array {
-  const array = new Uint8Array(string.length);
-  const { length } = string;
-  for (let i = 0; i < length; i++) {
-    array[i] = string.charCodeAt(i);
-  }
-  return array;
+export function utfEncodeString(string: string): Uint32Array {
+  // $FlowFixMe Flow's Uint32Array.from's type definition is wrong; first argument of mapFn will be string
+  return Uint32Array.from(string, toCodePoint);
```
Jessidhia: The correct definition, ignoring the 3rd argument, is

```ts
<T>(iterable: Iterable<T>, mapFn?: <T>(value: T, index: number) => number): UInt32Array
```

but the `mapFn` is defined to `(number) => number`.

PR #24, opened 2019-03-01
Display a full-width background under a selected item

fatfisz:
- `.FullSize` is the outer-most wrapper, it has a padding
- `.OverflowHider` is inside `.FullSize`, it has `overflow: hidden` and `position: relative`
- `.AutoSizerWrapper` and `.SelectedElementBackground` are inside `.OverflowHider`

We have to listen to scroll and save it in state. Then, if there is a selected element, we can calculate the top position of `.SelectedElementBackground`.

Nothing really changes for `.AutoSizerWrapper` besides the padding that is moved to another wrapper.

I understand that this might be a bit over the top, but I think it's the simplest solution based on multi-layering.

Also `:hover` is not really handled and would require switching to JS via mouseover/mouseout handling + an additional context value. I could do that too if this will help.

### DIFF
--- a/src/devtools/views/Element.css
+++ b/src/devtools/views/Element.css
@@ -14,8 +14,6 @@
 }
 
 .SelectedElement {
-  background-color: var(--color-tree-node-selected);
-
   /* Invert colors */
   --color-component-name: var(--color-component-name-inverted);
   --color-jsx-arrow-brackets: var(--color-jsx-arrow-brackets-inverted);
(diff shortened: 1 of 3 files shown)


fatfisz:
Also don't feel bad about closing this if you don't have time for review, I did it for fun.

bvaughn:
Hey @fatfisz! Thanks for the PR man!

I think the fix proposed by @GabbeV (8aa6f96) seems preferable since it doesn't require the imperative code, but this approach is a good fallback plan if that one proves to cause any unexpected problems.

I'm going to close this for now but I left a comment referring to this PR should we need to revert that change 😄

PR #33, opened 2019-03-17
Experimenting with portals [WIP]

bvaughn:
Rather than trying to synchronize between three separate React apps– maybe there could be a single app that is rendered within the current `ExtensionPanel` as a portal.

### TODO List
- [x] Figure out a good solution for style sharing between portals. (CSS vars are currently broken.)
- [x] Navigation between React websites currently breaks the portals/tabs.
- [x] Profiler doesn't work after navigating to a new page. (Records but then says "No profiling data has been recorded.")


shells/browser/shared/src/main.js
```diff
@@ -54,69 +63,85 @@ function createPanelIfReactLoaded() {
...
-        }
-      );
+          // TODO: When the user switches to the panel, check for an Elements tab selection.
+        });
+        panel.onHidden.addListener(() => {
+          // TODO: Stop highlighting and stuff.
+        });
+      });
 
       // TODO (profiling) Is there a way to detect profiling support and conditionally register this panel?
```
sompylasar: From a user's perspective it may be unclear why the panel is missing and what to do to enable profiling. Having the panel regardless gives an opportunity to explain this to the user on that panel.

shells/browser/shared/src/main.js
```diff
@@ -54,69 +63,85 @@ function createPanelIfReactLoaded() {
...
+            profilerPanel.render(renderRootToPortal, 'profiler');
+          }
+        });
+      });
 
-            settingsPanel = panel;
-          });
-        }
-      );
+      chrome.devtools.panels.create('⚛ Settings', '', 'panel.html', panel => {
```
sompylasar: Settings may be more efficient use of space as an overlay on top of an existing panel (on any of them) like the DevTools own settings.

bvaughn: That would require putting a toggle inside the panel which would take up precious space.

Anyway, feedback is usually appreciated but I'm not looking for UI/UX feedback on this PR. Only posted it in case anyone had suggestions on the shared styles issue, or the general portaling technique.

shells/browser/shared/src/main.js
bvaughn: This matches the current DevTools behavior.

sompylasar: > This matches the current DevTools behavior.

👍

shells/browser/shared/src/main.js
sompylasar: > That would require putting a toggle inside the panel which would take up precious space.

Yes, but this could be a small cog button or even smaller three-vertical-dots menu on the side.

> Anyway, feedback is usually appreciated but I'm not looking for UI/UX feedback on this PR. Only posted it in case anyone had suggestions on the shared styles issue, or the general portaling technique.

Yeah sorry for the distraction. Noticed this piece of code here so decided to report. There's no other way in GitHub to review/leave feedback on code or product decisions other than in a diff/PR or in an issue after the decisions have been finalized.

PR #38, opened 2019-04-01
Fix casing on store import

lucasecdb:
for some reason, this work on my mac, but when I tried to build on a linux machine it broke.


bvaughn:
Cool, thanks

PR #41, opened 2019-04-02
Avoid flashing "Loading..." in right pane

gaearon:
Maybe we could do it with Suspense but I couldn't quite figure out the right way to do it yet.

This avoids the flashing "Loading..." when you switch between items fast. (Instead, it opts for a temporary inconsistency between props and the title pane.)

### DIFF
--- a/src/devtools/views/Elements/SelectedElement.js
+++ b/src/devtools/views/Elements/SelectedElement.js
@@ -30,7 +30,9 @@ export default function SelectedElement(_: Props) {
   const element =
     selectedElementID !== null ? store.getElementByID(selectedElementID) : null;
 
-  const inspectedElement = useInspectedElement(selectedElementID);
+  const actualInspectedElement = useInspectedElement(selectedElementID);
+  // The UI lags behind to avoid flashing a loading state.
+  const inspectedElement = useLastNonNullValue(actualInspectedElement);
 
   const highlightElement = useCallback(() => {
     if (element !== null && selectedElementID !== null) {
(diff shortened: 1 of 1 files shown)


src/devtools/views/Elements/SelectedElement.js
```diff
@@ -30,7 +30,9 @@ export default function SelectedElement(_: Props) {
   const element =
     selectedElementID !== null ? store.getElementByID(selectedElementID) : null;
 
-  const inspectedElement = useInspectedElement(selectedElementID);
+  const actualInspectedElement = useInspectedElement(selectedElementID);
+  // The UI lags behind to avoid flashing a loading state.
```
bvaughn: This comment should probably be a little more detailed.

bvaughn:
> Maybe we could do it with Suspense but I couldn't quite figure out the right way to do it yet.

This would be a nicer long-term solution.

I hadn't started using suspense in DevTools yet when I wrote this part of the UI. I think I have a better handle on how I would use it for the inspected element now though.

gaearon:
Do you want to tweak as you see fit?

bvaughn:
Yeah sure thing. I was going to attach a patch via a comment for you to consider, but I could also just add it myself. :)

bvaughn:
Hm... this also tears between the title and the content area. 🤔

bvaughn:
I'm reconsidering this as a temporary solution, since it tears between the title above and the content. I think it's more acceptable for left and right panels to tear.

bvaughn:
Eh. It's fine. We can "fix it" better with #44

bvaughn:
I  think I'm going to revert this change for now. I just noticed a bug that was introduced as a result. (I added an intentional increased delay to `Agent` so that it would be more noticeable.)

bug-Kapture 2019-04-03 at 7 49 22

We should  fix the flashing behavior by using Suspense. (#44)

PR #45, opened 2019-04-02
Tab name: "Elements" -> "Components"

gaearon:
I think this disambiguates it more clearly with the Chrome Elements tab.

We also don't show all Elements anymore anyway.

Screen Shot 2019-04-02 at 7 55 21 PM

Second commit renames it in the code. Since label changes without naming changes can be confusing.


bvaughn:
Noiiice

Issue #52, opened 2019-04-03
Check if accessibility regressions exist compared to old DevTools


gaearon:
Before this becomes stable, we need to check if we are regressing accessibility on any important existing interactions.

At least, we should probably make the tree view focusable.

bvaughn:
One example that would be nice to improve (but I think might be trickier than you would assume) is keyboard focus and navigation within the Components tree.

bvaughn:
c60f991 and d1a5346 should significantly improve keyboard accessibility for the main "Components" tab. It's probably still pretty rough in the Profiler tab though.

How sure how to support keyboard navigation of the flamegraph. "chrome://tracing" uses WASD keys to zoom in and out, but its more of an analog zoom and I think that would require a significant change for DevTools to operate that way.

bvaughn:
Also commit https://react-devtools-experimental-ztikpbkpml.now.sh

AlmeroSteyn:
Hi, I love the look and feel of the new dev tools.

Been looking a bit at a11y regressions on the dev tools. Purely looking at regressions from the old dev tools I can currently find only one issue while testing with keyboard only as well as screen reader. I have not tested with other AT such as voice control or other input methods such as switches etc.

I tested with:
Windows 10
Firefox 66.0.3
Chrome 74.0.3729.131
Chrome Canary 76.0.3782.0
NVDA 2019.1.1

NVDA accounts for more than 30% of screen reader use.

Keyboard a11y has seen a great improvement in the new dev tools. Awesome work!

In the old dev tools I am unable to reach the Profiler sub tab or anything but the main bar buttons. This is also the reason for a single regression as almost the entire old devtools is keyboard inaccessible, hiding everything inside from inspection.

**The issue**

Button texts are no longer read out in NVDA screen reader. All controlling buttons are simply read out as `button`. In the old devtools all reachable buttons still have text read out by NVDA.

I see that the buttons carry an `aria-labelledby` property which points to the tooltips. However it is usually recommended that the elements decorated with the `id` values being referenced by `aria-labelledby` and `aria-describedby` exists in the rendered DOM if these properties are applied. 

The current state where they refer to `id` values even when they are not rendered will cause issues in some screen readers, NVDA being one. This has to do with how the accessibility tree is resolved by the screen reader and how it builds its own object model.

[...]

PR #53, opened 2019-04-03
Make tree reachable by focus

gaearon:
Related to https://github.com/bvaughn/react-devtools-experimental/issues/52.

This makes the tree focusable, and handles focus by selecting the first item if none are already selected.

Once you're inside the tree, the regular keyboard shortcuts take over.

I probably missed something but happy to work more on it if you can help me find the broken cases!

Screen Recording 2019-04-03 at 07 00 PM

### DIFF
--- a/src/devtools/views/Components/Tree.js
+++ b/src/devtools/views/Components/Tree.js
@@ -113,7 +113,7 @@ export default function Tree(props: Props) {
             <FixedSizeList
               className={styles.List}
               height={height}
-              innerElementType={innerElementType}
+              innerElementType={TreeWrapper}
               itemCount={numElements}
               itemData={itemData}
               itemSize={lineHeight}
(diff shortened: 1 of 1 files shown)


src/devtools/views/Components/Tree.js
```diff
@@ -113,7 +113,7 @@ export default function Tree(props: Props) {
             <FixedSizeList
               className={styles.List}
               height={height}
-              innerElementType={innerElementType}
+              innerElementType={TreeWrapper}
```
bvaughn: nit: I kind of liked the `innerElementType` name better. `TreeWrapper` to me indiates something that wraps this whole component (since its name is `Tree`) rather than something that goes inside of the list.

bvaughn:
I'm surprised this didn't cause problems, but I don't see any from my initial testing 😁 Nice!

Issue #25, opened 2019-03-01
Re-enable postMessage transferable for faster ArrayBuffer transfers


gaearon:
I got this on FB.com sandbox:

<img width="815" alt="screen shot 2019-03-01 at 1 15 24 pm" src="https://user-images.githubusercontent.com/810438/53640457-26dcbb00-3c24-11e9-828f-a987ffeec4da.png">

bvaughn:
Hey Dan. Any info about what you were doing when this happened? (Any breakpoints set or anything?) Can you reproduce it?

gaearon:
Nope, just happened randomly. :-(

bvaughn:
Damn!

I've seen this one before, but only when I had breakpoints set in Chrome.

I guess I should just comment out the transferrables approach for now since Chrome doesn't let me actually transfer the array buffer anyway. That sucks.

bvaughn:
How often are you able to trigger this? (If I thought I had a fix– which I don't, but if I did– would you be able to say with any confidence that it worked? or is it a rare error for you?)

bvaughn:
Regarding transferables, the [`postMessage` docs](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage) say:
> An optional array of Transferable objects to transfer ownership of. If the ownership of an object is transferred, it becomes unusable (neutered) in the context it was sent from and becomes available only to the worker it was sent to.

So it seems like either Chrome has a bug or (probably more likely) some part of the DevTools code is trying to touch the array buffer somehow _after_ it's been sent across the bridge. I'm not sure how this is happening though, because the flow of code is:

* [`renderer` passes the array to the `agent` and _immediately_ clears its reference](https://github.com/bvaughn/react-devtools-experimental/blob/8aa6f96279700ce75a1980f4cad97bafdc36eaa5/src/backend/renderer.js#L541-L542)
* [`Agent` passes the array on to the `Bridge`](https://github.com/bvaughn/react-devtools-experimental/blob/8aa6f96279700ce75a1980f4cad97bafdc36eaa5/src/backend/agent.js#L202-L208) but otherwise does nothing with it
* [`Bridge` passes it to the "wall", maybe after a small delay](https://github.com/bvaughn/react-devtools-experimental/blob/8aa6f96279700ce75a1980f4cad97bafdc36eaa5/src/bridge.js#L30-L50)– but it has not yet been serialized at this point so the delay should be okay
[...]

PR #57, opened 2019-04-04
Post message issue

DaniAcu:
I tried to solve issue #25 
- First I tried to add more structure to the data to prevent randomly cut with slice(0,3). I think that we didn't know what had on this scope when is running.
- Second I prevented the use on chrome for now.

Please if you have time check it. Thanks for the opportunity to contribute :smile:

### DIFF
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -47,7 +47,7 @@ export default class Bridge extends EventEmitter {
 
   _flush = () => {
     while (this._messageQueue.length) {
-      this._wall.send.apply(this._wall, ...this._messageQueue.splice(0, 1)[0]);
+      this._wall.send.apply(this._wall, Object.values(this._messageQueue.splice(0, 1)[0]));
     }
 
     if (this._timeoutID !== null) {
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -34,7 +34,7 @@ export default class Bridge extends EventEmitter {
       this._wall.send(event, payload, transferable);
       this._time = Date.now();
     } else {
-      this._messageQueue.push(event, payload, transferable);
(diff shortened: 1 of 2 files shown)


src/bridge.js
```diff
@@ -34,7 +34,7 @@ export default class Bridge extends EventEmitter {
       this._wall.send(event, payload, transferable);
       this._time = Date.now();
     } else {
-      this._messageQueue.push(event, payload, transferable);
+      this._messageQueue.push({ event, payload, transferable });
```
bvaughn: Can you explain why this change is actually important? Looks to me like all you're doing is wrapping args  in an object and unwrapping them below, but why?

DaniAcu:
When use send. Before you was using a slice(0,3). And you have error with ArrayBuffer, because you continue use the values on some part apparently . I thought that you could use a object to be sure that you are sending the correct information. But I have doubts about this. I continue think and maybe the best option is clean the buffer after that you use. Thoughts? I will try to clean the buffer later

bvaughn:
> When use send. Before you was using a slice(0,3).

`splice` was just being used to pull the next 3 values out of the queue, because [we always push three values onto the queue at a time](https://github.com/bvaughn/react-devtools-experimental/blob/a9fced9b0d4eab36f70da29a32fcaca3b2a37b6c/src/bridge.js#L37). It doesn't have anything to do with one of the values being an array.

It looks like all you're doing is putting all three values in a wrapper, and then pulling them back out again. I wouldn't expect this to have any impact (other than adding a little extra overhead).

DaniAcu:
Mmm you right. Yep, this change it's unnecessary. Let me think more on this tricky issue. Thanks for the feedback

Issue #60, opened 2019-04-04
"Show DOM node" doesn't work on Suspense in fallback mode


gaearon:
Shows up as 0x0 thing in the corner. While this might make sense as `findDOMNode` behavior, Ideally we'd highlight the fallback instead in this case.

bvaughn:
Since PR #61 fixes this (I think?) I'm going to assign this one to you.

Issue #43, opened 2019-04-02
Suspense Debugger


gaearon:
### MVP
- [x] Initial prototype with a tree and toggling.
- [x] Tree should reflect real suspense state.
- [ ] Make it less fragile.
  - [x] Does findDOMNode issue affect this?
- [ ] Ensure expensive tree construction doesn't run unless you're on the tab.
- [ ] Improve navigation so the user doesn't get lost.
  - [ ] Maybe flash nodes.
  - [ ] Maybe offer a "finder" tool.
- [ ] Sync selection with the Components tab.
- [ ] Make it look nice. @bvaughn 

### Future
- [ ] Consider a fuzzer.
- [ ] Considering filtering by viewport.
- [ ] Optimize the tree updates.
  - [ ] Consider listening to operations. Don't forget to handle ancestor reorder.

bvaughn:
This repository is being merged into the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)). As part of this, I am moving all issues to that repository as well and closing them here.

This issue has been relocated to:
https://github.com/facebook/react/issues/16489

PR #61, opened 2019-04-04
Allow to toggle Suspense in Components pane

gaearon:
Supersedes https://github.com/bvaughn/react-devtools-experimental/pull/36.

Fixes https://github.com/bvaughn/react-devtools-experimental/issues/60.

This is MVP support for toggling Suspense from the Components pane. We plan a separate tab in https://github.com/bvaughn/react-devtools-experimental/issues/43 but we'd want a way to quickly debug it anyway so seems like this is a good first step.

## Showing Suspense Status

Currently, Suspense shows Fiber internal state in the "state" section of the inspector (when Suspended), and nothing if unsuspended. This PR changes so that we show the current Suspense state:

Screen Recording 2019-04-04 at 03 29 PM

This works with older versions of React.

## Toggling Suspense Status

The inspected value turns into a checkbox if:

* You're using a version of React that includes https://github.com/facebook/react/pull/15232, **AND**
  - The component is not currently Suspended, **OR**
  - You've previously pressed the checkbox to force it to Suspend.

This means that you can toggle Suspense on and off. But if a component actually Suspends by itself, there's nothing to show in the "current" state. In that case the checkbox becomes a read-only label again until it un-Suspends.

Alternatively, we could separate the "current value" from the "override value". But I feel like this is sufficient and intuitive enough in practice after playing with it.

Screen Recording 2019-04-04 at 03 36 PM

## Implementation

The simplest implementation would be to always `setSuspenseHandler()` to the thing that checks IDs in the Set. However, I don't want to incur any Set checking overhead if none of the Suspenses were manually toggled. So I attach and detach the Set-checking handler on demand.

### DIFF
--- a/shells/dev/app/index.js
+++ b/shells/dev/app/index.js
@@ -32,7 +32,6 @@ function mountTestApp() {
   mountHelper(ElementTypes);
   mountHelper(EditableProps);
   mountHelper(DeeplyNestedComponents);
-  mountHelper(SuspenseTree);
 }
 
 function unmountTestApp() {
--- a/shells/dev/app/index.js
+++ b/shells/dev/app/index.js
@@ -32,6 +32,7 @@ function mountTestApp() {
   mountHelper(ElementTypes);
   mountHelper(EditableProps);
   mountHelper(DeeplyNestedComponents);
+  mountHelper(SuspenseTree);
 }
 
 function unmountTestApp() {
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -226,7 +226,7 @@ export function attach(
     setSuspenseHandler,
     scheduleUpdate,
   } = renderer;
-  const supportsEditingSuspense =
+  const supportsTogglingSuspense =
     typeof setSuspenseHandler === 'function' &&
     typeof scheduleUpdate === 'function';
 
@@ -1443,8 +1443,8 @@ export function attach(
       // Does the current renderer support editable function props?
       canEditFunctionProps: typeof overrideProps === 'function',
 
-      canEditSuspense:
-        supportsEditingSuspense &&
+      canToggleSuspense:
+        supportsTogglingSuspense &&
         // If it's showing the real content, we can always flip fallback.
         (!isTimedOutSuspense ||
           // If it's showing fallback because we previously forced it to,
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -1053,18 +1053,30 @@ export function attach(
     currentRootID = -1;
   }
 
-  // The naming is confusing.
-  // They deal with opaque nodes (fibers), not elements.
-  function getNativeFromReactElement(id: number) {
+  function findNativeByFiberID(id: number) {
     try {
-      const primaryFiber = getPrimaryFiber(idToFiberMap.get(id));
-      const hostInstance = renderer.findHostInstanceByFiber(primaryFiber);
-      return hostInstance;
+      const fiber = findCurrentFiberUsingSlowPath(idToFiberMap.get(id));
+      if (fiber === null) {
+        return null;
+      }
+      const isTimedOutSuspense =
+        fiber.tag === SuspenseComponent && fiber.memoizedState !== null;
+      if (!isTimedOutSuspense) {
+        // Normal case.
+        return renderer.findHostInstanceByFiber(fiber);
+      } else {
+        // A timed-out Suspense's findDOMNode is useless.
+        // Try our best to find the fallback directly.
+        const maybeFallbackFiber =
+          (fiber.child && fiber.child.sibling) || fiber;
+        return renderer.findHostInstanceByFiber(maybeFallbackFiber);
+      }
     } catch (err) {
       // The fiber might have unmounted by now.
       return null;
     }
   }
+
   function getFiberIDFromNative(
     hostInstance,
     findNearestUnfilteredAncestor = false
@@ -1731,7 +1743,7 @@ export function attach(
     getCommitDetails,
     getFiberIDFromNative,
     getInteractions,
-    getNativeFromReactElement,
+    findNativeByFiberID,
     getProfilingDataForDownload,
     getProfilingSummary,
     handleCommitFiberRoot,
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -199,6 +199,7 @@ export function attach(
     IndeterminateComponent,
     MemoComponent,
     SimpleMemoComponent,
+    SuspenseComponent,
   } = ReactTypeOfWork;
   const {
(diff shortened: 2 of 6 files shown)


gaearon:
I pushed another commit to fix https://github.com/bvaughn/react-devtools-experimental/issues/60.

Without it, it's kind of a pain to inspect large trees.

Screen Recording 2019-04-04 at 08 14 PM

shells/dev/app/SuspenseTree/index.js
```diff
@@ -0,0 +1,58 @@
...
+      </Suspense>
+      <LoadLater />
+    </div>
+  );
+}
+
+function LoadLater() {
+  const [loadChild, setLoadChild] = useState(0);
+  return (
+    <Suspense
```
bvaughn: Selecting this element in the DevTools tree and clicking the load/reload button causes an error:
```
Uncaught TypeError: Cannot read property 'parentID' of undefined
    at _loop (devtools.js:13578)
    at Bridge.<anonymous> (devtools.js:13649)
    at Bridge.emit (devtools.js:1621)
    at devtools.js:12785
    at devtools.js:21985
```

src/backend/renderer.js
```diff
@@ -1414,6 +1425,14 @@ export function attach(
       // Does the current renderer support editable function props?
       canEditFunctionProps: typeof overrideProps === 'function',
 
+      canEditSuspense:
```
bvaughn: naming nit `canToggleSuspense` ?

src/backend/renderer.js
```diff
@@ -219,7 +220,15 @@ export function attach(
...
   } = ReactSymbols;
 
-  const { overrideHookState, overrideProps } = renderer;
+  const {
+    overrideHookState,
+    overrideProps,
+    setSuspenseHandler,
+    scheduleUpdate,
+  } = renderer;
+  const supportsEditingSuspense =
```
bvaughn: naming nit `supportsTogglingSuspense` ?

src/backend/renderer.js
```diff
@@ -1661,19 +1692,59 @@ export function attach(
...
+  let forceFallbackForSuspenseIDs = new Set();
+  function shouldSuspendFiberAccordingToSet(fiber) {
+    const id = getFiberID(getPrimaryFiber(((fiber: any): Fiber)));
+    return forceFallbackForSuspenseIDs.has(id);
+  }
+
+  function overrideSuspense(id, forceFallback) {
+    if (
+      typeof setSuspenseHandler !== 'function' ||
+      typeof scheduleUpdate !== 'function'
```
bvaughn: nit could we just check `!supportsEditingSuspense` here?

shells/dev/app/index.js
```diff
@@ -32,6 +33,7 @@ function mountTestApp() {
   mountHelper(ElementTypes);
   mountHelper(EditableProps);
   mountHelper(DeeplyNestedComponents);
+  mountHelper(SuspenseTree);
```
bvaughn: nit Move this above `DeeplyNestedComponents` so we can test it without as much scrolling 😆

src/backend/renderer.js
```diff
@@ -1661,19 +1692,59 @@ export function attach(
...
+      }
+    } else {
+      forceFallbackForSuspenseIDs.delete(id);
+      if (forceFallbackForSuspenseIDs.size === 0) {
+        // Last override is gone. Switch React back to fast path.
+        setSuspenseHandler(shouldSuspendFiberAlwaysFalse);
+      }
+    }
+    const fiber = idToFiberMap.get(id);
+    scheduleUpdate(fiber);
```
bvaughn: Do we need to call this with the _current_ Fiber or is it robust enough to take either one?

src/devtools/views/Components/SelectedElement.js
```diff
@@ -170,11 +183,21 @@ function InspectedElementView({
...
+          }}
+          overrideValueFn={overrideSuspenseFn}
+        />
+      ) : (
+        <InspectedElementTree
+          label="state"
+          data={state}
+          overrideValueFn={overrideStateFn}
+        />
+      )}
```
bvaughn: I hope we don't end up needing more branching like this 😅

shells/dev/app/SuspenseTree/index.js
bvaughn: Actually this happens when you toggle via the inspector too (makes sense). Seems like a more important thing to fix.

toggle-bug-Kapture 2019-04-04 at 13 45 36

src/backend/renderer.js
gaearon: Flow wouldn’t be happy

bvaughn: Oh, right. lol. Ok

src/backend/renderer.js
gaearon: Either one would work.

bvaughn: That makes sense. Just wanted to confirm.

src/devtools/views/Components/SelectedElement.js
gaearon: Can refactor if we do. :-) I haven’t seen other cases but they might come up if we add `memoizedState` to other types.

shells/dev/app/index.js
gaearon: I figured maybe you test some other case more often. But fair re: DeeplyNested ones. :-)

bvaughn:
This looks good to me. Shall we merge it? 😄

gaearon:
Still need to fix the suspense fragment behavior. It doesn’t always show the right thing in the tree. Not fault of this PR but I’d like to fix them together.

gaearon:
I'll get this in, but I'll keep working on the actual fix for Suspense fragment stuff separately. Because the problem already exists in master anyway.

src/backend/renderer.js
```diff
@@ -1045,18 +1053,30 @@ export function attach(
...
+        fiber.tag === SuspenseComponent && fiber.memoizedState !== null;
+      if (!isTimedOutSuspense) {
+        // Normal case.
+        return renderer.findHostInstanceByFiber(fiber);
+      } else {
+        // A timed-out Suspense's findDOMNode is useless.
+        // Try our best to find the fallback directly.
+        const maybeFallbackFiber =
+          (fiber.child && fiber.child.sibling) || fiber;
+        return renderer.findHostInstanceByFiber(maybeFallbackFiber);
```
sophiebits: Should this logic to skip a timed-out Suspense's primary children live in React itself, in findCurrentHostFiber? I don't think it makes a lot of sense for this to live in devtools.

If you findDOMNode from outside a Suspense, you shouldn't get a node back from the primary children. (I think?)

cc @acdlite

PR #62, opened 2019-04-04
Move selection when typing in search field

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/26.

I noticed search often selects a random item.

But looking closer, it was because of a partial match:

This changes the algorithm to also change the selection if you continue typing in the search field. I think this feels better. It also matches what Chrome does in text search more closely. (Although not quite — but implementing 1:1 mapping seems like too much complexity right now.)

### DIFF
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -195,6 +195,7 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
   } = state;
 
   const prevSearchIndex = searchIndex;
+  const prevSearchText = searchText;
   const numPrevSearchResults = searchResults.length;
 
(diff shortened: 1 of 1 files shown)


src/devtools/views/Components/TreeContext.js
```diff
@@ -307,8 +308,8 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
     }
   }
 
-  // Changes in search index should override the selected element.
-  if (searchIndex !== prevSearchIndex) {
+  // Changes in search index or typing should override the selected element.
+  if (searchIndex !== prevSearchIndex || searchText.indexOf(prevSearchText) === 0) {
```
bvaughn: Nice! Yeah this makes sense.

PR #63, opened 2019-04-04
Match search at word boundaries

gaearon:
This makes searching a bit more robust.

Previously, `item` would match both `*Item*`, `List*Item*`, and `Iniv*iteM*om`.

I've changed regex to only consider a match if the first query letter is either at the beginning of the component name (such as the case with `Item`), or if it is at the word boundary (like in `ListItem`). Matches in the middle of the name where the first letter appears as lowercase are ignored.

This prevents matching `c` to `InteractionTracking`, for example:

Screen Recording 2019-04-04 at 04 25 PM

## Implementation

See comments. To make this work, I had to make the regex case-sensitive. Instead, I expand the casings directly when building regex, and add a special case for the first letter.

`Listitem` would still match `ListItem`. Only the first letter in the target text is treated contextually.

### DIFF
--- a/src/devtools/views/Components/SearchInput.js
+++ b/src/devtools/views/Components/SearchInput.js
@@ -95,7 +95,7 @@ export default function SearchInput(props: Props) {
         onChange={handleTextChange}
         onKeyDown={handleKeyDown}
         onKeyPress={handleInputKeyPress}
-        placeholder="Search (text or /regex/)"
+        placeholder="Search"
         ref={inputRef}
         value={searchText}
       />
--- a/src/devtools/views/Components/SearchInput.js
+++ b/src/devtools/views/Components/SearchInput.js
@@ -95,7 +95,7 @@ export default function SearchInput(props: Props) {
         onChange={handleTextChange}
         onKeyDown={handleKeyDown}
         onKeyPress={handleInputKeyPress}
-        placeholder="Search"
+        placeholder="Search (text or /regex/)"
         ref={inputRef}
         value={searchText}
       />
--- a/src/devtools/views/utils.js
+++ b/src/devtools/views/utils.js
@@ -6,6 +6,23 @@ import { meta } from '../../hydration';
 import type { HooksTree } from 'src/backend/types';
 
 export function createRegExp(string: string): RegExp {
+  // Allow /regex/ syntax with optional last /
+  if (string[0] === '/') {
+    // Cut off first slash
+    string = string.substring(1);
+    // Cut off last slash, but only if it's there
+    if (string[string.length - 1] === '/') {
+      string = string.substring(0, string.length - 1);
+    }
+    try {
+      return new RegExp(string, 'i');
+    } catch (err) {
+      // Bad regex. Make it not match anything.
+      // TODO: maybe warn in console?
+      return new RegExp('.^');
+    }
+  }
+
   function isLetter(char: string) {
     return char.toLowerCase() !== char.toUpperCase();
   }
--- a/src/devtools/views/utils.js
+++ b/src/devtools/views/utils.js
@@ -6,7 +6,34 @@ import { meta } from '../../hydration';
 import type { HooksTree } from 'src/backend/types';
 
 export function createRegExp(string: string): RegExp {
-  return new RegExp(escapeStringRegExp(string), 'i');
+  // 'item' should match 'Item' and 'ListItem', but not 'InviteMom'.
+  // To do this, we'll slice off 'tem' and check first letter separately.
+  const escaped = escapeStringRegExp(string);
+  const firstLetter = escaped[0];
+  let restRegex = '';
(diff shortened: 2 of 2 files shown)


gaearon:
Hmm I guess this breaks the regex support.

bvaughn:
I like the direction you're going in from a UX perspective, but I wonder how much slower it would make search for large trees.

gaearon:
Or was it already broken? Typing `/item/` on master doesn't work either.

gaearon:
I can try on Ads Manager.

bvaughn:
I didn't actively intended to support regex input previously. (The code called `escapeStringRegExp` to escape special regex characters)

bvaughn:
> I can try on Ads Manager.

Sure. I  guess we can just make  an educated guess for  now and if we need to roll it back later, we can.

gaearon:
I'll remove `(text or /regex/)` from label if we don't intend to support?

gaearon:
Or I can add support as it's pretty simple.

bvaughn:
Yeah I think actually supporting would be nice. I guess at one point (when I wrote the tooltip) I meant to support both but then when I wrote the util code I forgot 😅

bvaughn:
I wonder if we could do something like this?

1. If the string starts with "/" - use it as a regex. (Auto-add the closing "/" to avoid breaking things while they're typing.) ((How should we handle invalid regexes?))
1. If the string is all lowercase letters, do case insensitive search.
1. If the string contains any uppercase letters, do case sensitive search. (This would let "Item" match  "ListItem" but not "InviteMom".)

gaearon:
>If the string is all lowercase letters, do case insensitive search.

I don't think I agree with this part. I think most people will always write "item" and will never think that casing matters.

I can add regex support though.

gaearon:
Added regex support. 👍

bvaughn:
> I don't think I agree with this part. I think most people will always write "item" and will never think that casing of the query matters.

Why would this matter then? If they write "item" then I think the current, case insensitive matching approach _should_ just work like it has.

The _change_ I proposed here would only be if they wrote "Item"

gaearon:
>Why would this matter then? If they write "item" then I think the current, case insensitive matching approach should just work like it has (e.g. match "ListItem", "InviteMom", etc)

But that was the thing I wanted to fix 😛 As a user I always want the non-InviteMom-matching behavior, and I'd never think to type `item` from upper case to enable that.

gaearon:
In other words: I think the vast majority of users (at least except beginners) will always type `button` rather than `Button`, etc. So they won't get my fix if we only enable it for search strings starting with capital letters.

bvaughn:
Interesting. I'm not sure that's a safe assumption for everyone, based on the fact that you and I want different behavior for the base case 😄

gaearon:
Would it help if I illustrated it with gifs showing old/new behavior? I think we might be overfocusing on this specific example and it's easier to see the pattern on real examples.

It would definitely help me if you could show concrete examples of where you'd prefer the current behavior. I can't think of one but I must be missing something.

bvaughn:
The current behavior was written to mirror the browser's find-in-page functionality. This isn't exactly an apples-to-apples comparison, since you're searching arbitrary text in the page rather than component names, but I think it's reasonable prior art given that this extension is also running in the browser.

My concern is that the behavior you're describing might not be _intuitive_ in all cases. This is just a subjective "gut feeling" really.

bvaughn:
I don't feel _very strongly_ about this. It's just a gut feeling. If you're confident that your approach isn't significantly worse (performance) I'll be happy to give it a try and see if we get any negative feedback :)

gaearon:
>The current behavior was written to mirror the browser's find-in-page functionality.

I guess I'm thinking of IDE fuzzy search as a closer analog here, at least in terms of the goal ("find a named item" vs "find a text match"). I see your point about matching the page text behavior but I'm not sure it is as valuable in this case — or at least I think partial matches are actively harming the experience because they keep matching on stuff that I don't want to see. Since regex is now supported, you can still do partial matches with a regex and I think that might be sufficient.

gaearon:
But also, I just struggle to think of a use case where I'd want to type `utton`, `tton`, `ton`, `on`, or `n` instead of `Button` (or `Butto`, etc) when searching for a component name.

I guess it could happen for something like `Breadcrumb` (where I could type `crumb`) but then it seems easy to remember once it doesn't work.

bvaughn:
Yeah. I have a nagging suspicion that I do _sometimes_ search that way (word endings) if my memory is fuzzy, but... maybe you're right that it's not very common and easy to work around.

When I have some cycles, I'll review this code (assuming it's ready?). I'm willing to give this UX a shot :)

gaearon:
Should be ready! Also ran it on some heavy Ads Manager screens — don't see perf drops.

bvaughn:
This feels nice. I dig it.

Issue #42, opened 2019-04-02
Cannot read property 'length' of undefined


gaearon:
<img width="445" alt="Screen Shot 2019-04-02 at 7 29 38 PM" src="https://user-images.githubusercontent.com/810438/55426975-f4551380-557d-11e9-9a49-d17e622e654e.png">

Happens because there's no `subHooks` property:

<img width="320" alt="Screen Shot 2019-04-02 at 7 30 52 PM" src="https://user-images.githubusercontent.com/810438/55426993-fae38b00-557d-11e9-8fb7-4cfb511b2656.png">

Steps to reproduce:

1. Load FB page
2. Search for "Suspense" (Note that this will immediately select an unrelated node — maybe same issue as https://github.com/bvaughn/react-devtools-experimental/issues/26)
3. Press Enter a few times

You'll see the crash.

gaearon:
This seems to happen because of

```js
const LEVEL_THRESHOLD = 6;
```

Do we never expect objects to be that deep? Is more depth meant to be requested on demand?

<img width="749" alt="Screen Shot 2019-04-04 at 7 10 45 PM" src="https://user-images.githubusercontent.com/810438/55578343-95be9f80-570d-11e9-8cde-f74051f8e495.png">

PR #64, opened 2019-04-04
Don't crash for deeply nested Hooks

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/42.

This happened because the code didn't handle the dehydrated case, and the hydration level constant is set at `6`. (Which is a few levels of custom Hooks + a few levels of `subHooks` arrays.)

The fix includes a few stress tests. The nested deep Hook one used to crash, but doesn't anymore.

Screen Shot 2019-04-04 at 7 24 47 PM

It's not ideal that we don't let you inspect deeper. But this can be fixed later with a UI that loads them on demand. I added todos for that.

### DIFF
--- a/src/devtools/views/Components/HooksTree.js
+++ b/src/devtools/views/Components/HooksTree.js
@@ -9,6 +9,7 @@ import EditableValue from './EditableValue';
 import KeyValue from './KeyValue';
 import { serializeHooksForCopy } from '../utils';
 import styles from './HooksTree.css';
+import { meta } from '../../../hydration';
 
 import type { HooksNode, HooksTree } from 'src/backend/types';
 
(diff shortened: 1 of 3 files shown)


src/devtools/views/Components/HooksTree.js
```diff
@@ -71,6 +72,11 @@ type HookViewProps = {|
 
 function HookView({ canEditHooks, hook, id, path = [] }: HookViewProps) {
   const { name, id: hookID, isStateEditable, subHooks, value } = hook;
+  if (hook.hasOwnProperty(meta.inspected)) {
+    // This Hook is too deep and hasn't been hydrated.
+    // TODO: show UI to load its data.
+    return null;
+  }
```
bvaughn: This early return needs to happen after `useContext` below.

I think we should also render _something_ to indicate that we've truncated (even if just "...").

I also think we'll need to handle this general case for deep props/state/context.

(I'll address the lint thing as part of adding a temporary truncation indicator.)

Screen Shot 2019-04-04 at 2 08 25 PM

PR #65, opened 2019-04-04
Bump font size of owner stack items

gaearon:
Not sure if you intentionally made it smaller or it was accidental.

With Comfortable layout, they're currently 11px — which IMO is a bit too small, at least for Comfortable. This adds the closest constant I found. I don't mind them being a bit smaller than that, but IMO they're way too small right now, at least for a Comfortable mode, to the point of being hard to read.

Thanks for considering.

### DIFF
--- a/src/devtools/views/Components/SelectedElement.css
+++ b/src/devtools/views/Components/SelectedElement.css
@@ -32,6 +32,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   color: var(--color-component-name);
+  font-size: var(--font-size-sans-normal);
 }
 .Component:before,
 .Owner:before {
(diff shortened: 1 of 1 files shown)


src/devtools/views/Components/SelectedElement.css
```diff
@@ -32,6 +32,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   color: var(--color-component-name);
+  font-size: var(--font-size-sans-normal);
```
bvaughn: Whoops, looks like an oversight. These should be both:
```css
    font-family: var(--font-family-monospace);
    font-size: var(--font-size-monospace-normal);
```

PR #66, opened 2019-04-04
Stop mouseup propagation while inspecting

gaearon:
I noticed that in some places on FB, "selecting" an element inside a context menu will dismiss the menu. While we stop `click` and `mousedown` propagation, we didn't stop `mouseup`. Our handlers listen to it.

This makes clicking the page to choose a DOM element a bit less likely to blow away the thing you're trying to inspect — by preventing `mouseup` propagation too until your click is done.


bvaughn:
Nice find.

Issue #68, opened 2019-04-04
Uncaught TypeError: Cannot read property 'parentID' of undefined


gaearon:
With this case:

```js
import React, { Suspense, useState } from 'react';

function SuspenseTree() {
  return (
    <>
      <h1>Suspense</h1>
      <Suspense fallback={<h2>Loading outer</h2>}>
        <Parent />
      </Suspense>
    </>
  );
}

function Never() {
  throw new Promise(resolve => {});
}

function Parent() {
  return (
    <div>
      <Suspense fallback={<h3>Loading inner 1</h3>}>
        <Child>Hello</Child>
      </Suspense>
      <Suspense fallback={<h3>Loading inner 2</h3>}>
        <Child>World</Child>
      </Suspense>
      <Suspense fallback={<h3>This will never load</h3>}>
        <Never />
      </Suspense>
      <LoadLater />
    </div>
  );
}

function LoadLater() {
  const [loadChild, setLoadChild] = useState(0);
  return (
    <Suspense
      fallback={<h4 onClick={() => setLoadChild(true)}>Click to load</h4>}
    >
      {loadChild ? (
        <h4 onClick={() => setLoadChild(false)}>
          Loaded! Click to suspend again.
        </h4>
      ) : (
        <Never />
      )}
    </Suspense>
  );
}

function Child({ children }) {
  return <p>{children}</p>;
}
```

Clicking on "load" / "suspend" / "load" three times causes an error:

```
Uncaught TypeError: Cannot read property 'parentID' of undefined
    at _loop (devtools.js:13988)
    at Bridge.<anonymous> (devtools.js:14059)
    at Bridge.emit (devtools.js:1623)
    at devtools.js:13195
    at devtools.js:22390
_loop @ devtools.js:13988
(anonymous) @ devtools.js:14059
emit @ devtools.js:1623
(anonymous) @ devtools.js:13195
(anonymous) @ devtools.js:22390
postMessage (async)
send @ backend.js:14977
send @ backend.js:14504
(anonymous) @ backend.js:12211
(anonymous) @ devtools.js:21895
emit @ devtools.js:21894
flushPendingEvents @ backend.js:12983
handleCommitFiberRoot @ backend.js:13380
onCommitFiberRoot @ devtools.js:21933
(anonymous) @ app.js:3565
(anonymous) @ app.js:3559
onCommitRoot @ app.js:3567
commitRoot @ app.js:5089
completeRoot @ app.js:5418
[...]

PR #71, opened 2019-04-04
Fix Suspense fragment edge cases

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/68.

You can verify with the fixture. (It would be nice to expand it or make a fuzzer in the future.)
I don't know if the fix is 100% correct, but it should be more correct than master.

Screen Recording 2019-04-05 at 12 29 am

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -22,6 +22,7 @@ import {
   TREE_OPERATION_ADD,
   TREE_OPERATION_REMOVE,
   TREE_OPERATION_RESET_CHILDREN,
+  TREE_OPERATION_RECURSIVE_REMOVE_CHILDREN,
   TREE_OPERATION_UPDATE_TREE_BASE_DURATION,
 } from '../constants';
 import { getUID } from '../utils';
@@ -627,7 +628,7 @@ export function attach(
     pendingOperations = new Uint32Array(0);
   }
 
-  function enqueueMount(fiber: Fiber, parentFiber: Fiber | null) {
+  function recordMount(fiber: Fiber, parentFiber: Fiber | null) {
     const isRoot = fiber.tag === HostRoot;
     const id = getFiberID(getPrimaryFiber(fiber));
 
@@ -713,7 +714,7 @@ export function attach(
     }
   }
 
-  function enqueueUnmount(fiber) {
+  function recordUnmount(fiber) {
     const isRoot = fiber.tag === HostRoot;
     const primaryFiber = getPrimaryFiber(fiber);
     if (!fiberToIDMap.has(primaryFiber)) {
@@ -753,18 +754,27 @@ export function attach(
     }
   }
 
-  function mountFiber(
+  function recordRecursiveRemoveChildren(fiber) {
+    const primaryFiber = getPrimaryFiber(fiber);
+    const id = getFiberID(primaryFiber);
+    const operation = new Uint32Array(2);
+    operation[0] = TREE_OPERATION_RECURSIVE_REMOVE_CHILDREN;
+    operation[1] = id;
+    addOperation(operation, false);
+  }
+
+  function mountFiberRecursively(
     fiber: Fiber,
     parentFiber: Fiber | null,
     traverseSiblings = false
   ) {
     if (__DEBUG__) {
-      debug('mountFiber()', fiber, parentFiber);
+      debug('mountFiberRecursively()', fiber, parentFiber);
     }
 
-    const shouldEnqueueMount = !shouldFilterFiber(fiber);
-    if (shouldEnqueueMount) {
-      enqueueMount(fiber, parentFiber);
+    const shouldInclude = !shouldFilterFiber(fiber);
+    if (shouldInclude) {
+      recordMount(fiber, parentFiber);
     }
 
     const isTimedOutSuspense =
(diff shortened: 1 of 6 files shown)


src/backend/renderer.js
```diff
@@ -707,6 +707,16 @@ export function attach(
   function enqueueUnmount(fiber) {
     const isRoot = fiber.tag === HostRoot;
     const primaryFiber = getPrimaryFiber(fiber);
+    if (!fiberToIDMap.has(primaryFiber)) {
```
gaearon: We need this because React informs us about _every_ unmount, including "not-conceptually-primary" Suspense children tree. However, our maps only "remember" things in the "conceptually-primary" tree.

src/backend/renderer.js
```diff
@@ -743,8 +753,24 @@ export function attach(
...
+      fiber.tag === ReactTypeOfWork.SuspenseComponent &&
+      fiber.memoizedState !== null;
+
+    if (isTimedOutSuspense) {
+      // Special case: if Suspense mounts in a timed-out state,
+      // get the fallback child from the inner fragment and mount
+      // it as if it was our own child. Updates handle this too.
+      const primaryChildFragment = fiber.child;
+      const fallbackChildFragment = primaryChildFragment.sibling;
+      const fallbackChild = fallbackChildFragment.child;
```
acdlite: This part looks correct. It corresponds to this part of `updateSuspenseComponent`:

https://github.com/facebook/react/blob/49595e921dcb748c0422bc7aaadf2d3288ff1c46/packages/react-reconciler/src/ReactFiberBeginWork.js#L1434-L1465

gaearon:
I think there are still missing cases in the original logic that I haven't covered here.

We need to check what happens when:

* we mount primary
* we mount fallback
* we update primary -> fallback
* we update primary -> primary
* we update fallback -> primary
* we update fallback -> fallback

My impression is we only handle a subset of these today.

gaearon:
I'll try to write a more comprehensive fix tomorrow.

src/backend/renderer.js
```diff
@@ -713,9 +714,19 @@ export function attach(
     }
   }
 
-  function enqueueUnmount(fiber) {
+  function recordUnmount(fiber) {
```
bvaughn: `function recordUnmount(fiber: Fiber) {`

src/backend/renderer.js
```diff
@@ -832,109 +888,144 @@ export function attach(
...
-      if (shouldEnqueueUpdate) {
-        enqueueUpdateIfNecessary(nextFiber, hasChildOrderChanged);
+      // If we have no more children, but used to, they don't line up.
+      if (!hasChildOrderChanged && prevChildAtSameIndex != null) {
+        hasChildOrderChanged = true;
       }
     }
+
+    if (shouldInclude) {
+      maybeRecordUpdate(nextFiber, hasChildOrderChanged);
```
bvaughn: nit the combination of `shouldInclude` and `maybeRecordUpdate` confused me at first glance. Maybe `shouldInclude` could be a little more obvious at a glance if it had a more detailed name? e.g. `shouldIncludeInTree`?

src/devtools/store.js
```diff
@@ -572,7 +573,52 @@ export default class Store extends EventEmitter {
...
+                '. This is a bug in React DevTools.'
+            );
+          }
+
+          i = i + 2;
+
+          let justRemovedIDs = [];
+          const recursivelyRemove = id => {
+            justRemovedIDs.push(id);
+            const el = this._idToElement.get(id);
```
bvaughn: tiny nit subjective dislike of abbreviated names

src/devtools/store.js
```diff
@@ -3,6 +3,7 @@
 import EventEmitter from 'events';
 import {
   TREE_OPERATION_ADD,
+  TREE_OPERATION_CLEAR_CHILDREN,
```
bvaughn: super nit "clear" is another way of saying "remove" I guess, but I wonder if a name like `TREE_OPERATION_RECURSIVE_REMOVE_CHILDREN` would be "clearer" heh heh

src/devtools/store.js
```diff
@@ -572,7 +573,52 @@ export default class Store extends EventEmitter {
...
+          let startIndex = oldRemovedElementIDs.length;
+          for (let j = 0; j < justRemovedIDs.length; j++) {
+            removedElementIDs[startIndex + j] = oldRemovedElementIDs[j];
+          }
+
+          parentElement = ((this._idToElement.get(id): any): Element);
+          parentElement.children.forEach(recursivelyRemove);
+          parentElement.children = [];
+          weightDelta = -parentElement.weight + 1;
+          break;
```
bvaughn: Kind of a bummer that this requires so much overlap with the `TREE_OPERATION_REMOVE` operation.

src/devtools/store.js
```diff
@@ -614,11 +660,12 @@ export default class Store extends EventEmitter {
           }
 
           // Track removed items so search results can be updated
-          const oldRemovededElementIDs = removedElementIDs;
```
bvaughn: 😆

src/devtools/store.js
```diff
@@ -655,6 +702,10 @@ export default class Store extends EventEmitter {
           i = i + 3;
           break;
         default:
+          // ----------
+          // TODO: I get here with 0 sometimes after toggling suspense :-(
+          // ----------
```
bvaughn: "with 0" meaning...  the next operation in the array? Yikes. That's concerning.

bvaughn: Could this be because you didn't actually add the `TREE_OPERATION_CLEAR_CHILDREN` constant? so when you push the undefined value, it's just defaulting to 0?

bvaughn: (in which case I kind of wonder how the above `TREE_OPERATION_CLEAR_CHILDREN` case would ever be hit...)

src/backend/renderer.js
bvaughn: Bump~

bvaughn:
With regard to 7a03c5a, I assume you just want to keep commits separate, but I assume you'll also update `CommitTreeBuilder` to be aware of the new operation type?

gaearon:
I did that in https://github.com/bvaughn/react-devtools-experimental/pull/71/commits/80d9d8d841d02689d4c75360e2b42b543ab3477e :-)

Issue #6, opened 2019-02-22
Tree sometimes receives duplicate add-item operations


bvaughn:
An inline comment I wrote a few weeks ago indicates that this might have something to do with `React.lazy` or `React.Suspense` (fallback elements). Would be nice to track down and fix the cause of this so we could remove [these](https://github.com/bvaughn/react-devtools-experimental/blob/0c346ceea8650219b1dbb43d7bb0ebbe5732b318/src/devtools/store.js#L243-L247) [checks](https://github.com/bvaughn/react-devtools-experimental/blob/0c346ceea8650219b1dbb43d7bb0ebbe5732b318/src/devtools/store.js#L292-L296) from the `Store`.

PR #75, opened 2019-04-05
Fix double-adding fibers when traversing

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/6.

This used to happen because when converting a loop version to the recursive version of `mountFiber`, we always went into `fiber.sibling`. But we should have only done that for the inner iterations — not the topmost one.

Adds a test that demonstrates the issue on master.

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -743,23 +743,26 @@ export function attach(
     }
   }
 
-  function mountFiber(fiber: Fiber, parentFiber: Fiber | null) {
+  function mountFiber(
+    fiber: Fiber,
+    parentFiber: Fiber | null,
+    traverseSiblings = false
+  ) {
     if (__DEBUG__) {
       debug('mountFiber()', fiber, parentFiber);
     }
 
     const shouldEnqueueMount = !shouldFilterFiber(fiber);
-
     if (shouldEnqueueMount) {
       enqueueMount(fiber, parentFiber);
     }
 
     if (fiber.child !== null) {
-      mountFiber(fiber.child, shouldEnqueueMount ? fiber : parentFiber);
+      mountFiber(fiber.child, shouldEnqueueMount ? fiber : parentFiber, true);
     }
 
-    if (fiber.sibling) {
-      mountFiber(fiber.sibling, parentFiber);
+    if (traverseSiblings && fiber.sibling !== null) {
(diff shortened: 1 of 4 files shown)


src/backend/renderer.js
```diff
@@ -743,23 +743,26 @@ export function attach(
     }
   }
 
-  function mountFiber(fiber: Fiber, parentFiber: Fiber | null) {
+  function mountFiber(
+    fiber: Fiber,
+    parentFiber: Fiber | null,
+    traverseSiblings = false
```
gaearon: I originally wanted to split into two functions but that felt a bit more confusing. I'm ok with this signature but I can split if you want.

src/backend/renderer.js
```diff
@@ -743,23 +743,26 @@ export function attach(
...
     if (fiber.child !== null) {
-      mountFiber(fiber.child, shouldEnqueueMount ? fiber : parentFiber);
+      mountFiber(fiber.child, shouldEnqueueMount ? fiber : parentFiber, true);
     }
 
-    if (fiber.sibling) {
-      mountFiber(fiber.sibling, parentFiber);
+    if (traverseSiblings && fiber.sibling !== null) {
+      mountFiber(fiber.sibling, parentFiber, true);
     }
```
bvaughn: Nice catch!

src/devtools/store.js
```diff
@@ -482,37 +482,39 @@ export default class Store extends EventEmitter {
 
           i = i + 3;
 
+          if (__DEBUG__) {
+            if (this._idToElement.has(id)) {
+              console.error(
+                'Store already contains fiber ' + id + '. This is a bug.'
+              );
```
bvaughn: We should probably just throw here, since this is no longer expected to ever actually happen.

src/devtools/store.js
```diff
@@ -545,40 +547,42 @@ export default class Store extends EventEmitter {
...
           }
           break;
         case TREE_OPERATION_REMOVE:
           id = ((operations[i + 1]: any): number);
 
+          if (__DEBUG__) {
+            if (!this._idToElement.has(id)) {
+              console.error(
+                'Store does not contain fiber ' + id + '. This is a bug.'
+              );
```
bvaughn: Ditto for throwing instead of logging? Particularly since this would error right below when we try to read `element.parentID`

PR #79, opened 2019-04-06
[WIP] Use Suspense in the right pane

gaearon:
This is still janky when doing quick keyboard selection by long-pressing "down" arrow. I'm not sure why and how I should be modeling this. Sending a PR to kick off a discussion.


gaearon:
@bvaughn made a proper version of this. Closing.

bvaughn:
Thanks for the triage

PR #81, opened 2019-04-06
Show owner in DOM highilght overlay

gaearon:
This shows the owner name in DOM highlighter when the owner information is available.

I think this makes inspecting unfamiliar hierarchies a bit easier.

Screen Recording 2019-04-06 at 05 04 pm

### DIFF
--- a/src/backend/agent.js
+++ b/src/backend/agent.js
@@ -443,6 +443,7 @@ export default class Agent extends EventEmitter {
 
     const target = ((event.target: any): HTMLElement);
 
-    showOverlay(target, target.tagName.toLowerCase());
+    // Pass an empty string so it can infer name from tag and owner.
+    showOverlay(target, '');
   };
 }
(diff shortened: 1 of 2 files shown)


bvaughn:
Oooh I like it!

src/backend/agent.js
```diff
@@ -443,6 +443,7 @@ export default class Agent extends EventEmitter {
 
     const target = ((event.target: any): HTMLElement);
 
-    showOverlay(target, target.tagName.toLowerCase());
+    // Pass an empty string so it can infer name from tag and owner.
+    showOverlay(target, '');
```
bvaughn: Why empty string instead of just undefined?

src/backend/views/Overlay.js
```diff
@@ -145,6 +153,37 @@ export default class Overlay {
...
+function getFiber(node) {
+  if (
+    lastFoundInternalKey !== null &&
+    node.hasOwnProperty(lastFoundInternalKey)
+  ) {
+    return (node: any)[lastFoundInternalKey];
+  }
+  let internalKey = Object.keys(node).find(
+    key => key.indexOf('__reactInternalInstance') === 0
+  );
```
bvaughn: 😆

PR #82, opened 2019-04-06
RFC (not for merge): Show owner tree by default

gaearon:
### Note: Ignore the code, it's super hacky and doesn't really work except for a five second demo

Currently we have two places where owner tree shows up:

* Right pane
* In "owner mode" (which you enter with a double click)

I wonder if we could make it more prominent as it gives such a great bird eye's view of where you are, especially in giant trees.

The downside of the right pane is that it's buried under props and state so you often have to scroll down for it. The "owner stack" thing vertical position is always different so it's hard to train to see it quickly. Additionally, the "navigation" mostly happens on the left side already. I'm not proposing to remove the right pane but I wish there was a quicker way to find where I am.

The double-click "owner mode" is neat but I'd use it to *focus* on some part of the hierarchy. However, in large trees I often don't even know what I'm looking at. Is that the right or the wrong part?

**I wish I could quickly tell where the selected element roughly is in the tree without performing either scrolling or double-clicking (which moves me into a totally different mode)**.

One could argue we don't have space for breadcrumbs. But we do have the Search input. It's useful while you're searching — but when it's empty, it just takes up vertical space. I need to click to search either way, so it's not saving me much time by being large.

My proposal is to make some variation of the owner tree stack visible *by default* in the top bar. There would be a "search button" that moves you into the search mode (in which case you would see the current search UI). Exiting search mode would give you the breadcrumbs back. And the existing "owner mode" would just be a "focused" variation of the main mode where you only see a particular subtree.

This gif and PR is a very poor (literally, it will hang your browser in a few seconds) prototype of how that could feel. Notice that even though it's super janky (which we could fix), you get some sense of how the tree is structured from a single look.

Screen Recording 2019-04-06 at 05 33 pm

Is this a bad idea?

If there is interest, I can take a shot at implementing a real version.

### DIFF
--- a/shells/dev/app/ElementTypes/index.js
+++ b/shells/dev/app/ElementTypes/index.js
@@ -10,7 +10,7 @@ import React, {
   unstable_ConcurrentMode as ConcurrentMode,
   Fragment,
   // $FlowFixMe Flow doesn't know about the Profiler import yet
-  unstable_Profiler as Profiler,
+  Profiler,
   StrictMode,
   Suspense,
 } from 'react';
--- a/src/devtools/views/Components/OwnersStack.js
+++ b/src/devtools/views/Components/OwnersStack.js
@@ -1,31 +1,38 @@
 // @flow
 
-import React, { useCallback, useContext } from 'react';
+import React, { useCallback, useContext, useEffect, useState, useRef } from 'react';
 import Button from '../Button';
-import ButtonIcon from '../ButtonIcon';
+import Icon from '../Icon';
 import { TreeContext } from './TreeContext';
-import { StoreContext } from '../context';
+import { BridgeContext, StoreContext } from '../context';
+import { hydrate } from 'src/hydration';
 
 import type { Element } from './types';
 
 import styles from './OwnersStack.css';
 
-export default function OwnerStack() {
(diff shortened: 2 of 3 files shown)


bvaughn:
Hmm... going to want to think about this a bit. I'm sure we could figure out a way to make the search/list toggle UX work, **but** I think in many apps, the approach you're showing might not end up working, since the owners list gets too big quickly and we actually end up needing to hide it in its own "..." menu (#7)

bvaughn:
I wonder if there's a way  to surface your position within the overall tree without running into the crappy horizontal-overflowing-toolbar issue mentioned in #7 ...

gaearon:
I'll need to look at problems caused by #7 in realistic trees. Wondering if there's any way we can rank "important" owners and show ... between them.

bvaughn:
I was kind of imagining we might tweak #7 solution and show as many as we can, then hide the rest behind "..." - but I'm not sure how performant that will end up being.

Not sure how we could rank them really. Naive initial guess is that closest owners are  more important / more relevant than distant  ones when you've selected a given node.

gaearon:
I was also thinking maybe owners with lots of children are more important. Because they're the point where the tree "branches out".

Issue #74, opened 2019-04-05
Throttle Profiler screenshots to avoid hurting perf for e.g. "scroll"


bvaughn:
Show a "throttled" placeholder or something in the UI

bvaughn:
d766d3b does a few things to hopefully alleviate this a bit:
1. Profiler screenshots are disabled by default.
1. Enabling them shows a message saying they'll be throttled to avoid harming performance.
1. During profiling, the `Store` throttles to ensure we don't ask for more than two screenshots per second. (Ideally I would not count the time spent waiting on the response but I think this is ok for now.)
1. The Profiler UI shows a placeholder  "_No screenshot available_" message for commits that are missing screenshots.

PR #83, opened 2019-04-06
Continuously update tree selection in DOM inspection mode

gaearon:
I think this helps explore *what exactly* you're trying to select, as well get a general sense of the hierarchy. To reduce the noise, I throttled the highlighting by 200ms. That also helps avoid highlighting too early when you're just moving the cursor from the panel into the app.


src/backend/agent.js
```diff
@@ -339,6 +342,7 @@ export default class Agent extends EventEmitter {
     window.addEventListener('mousedown', this._onMouseDown, true);
     window.addEventListener('mouseup', this._onMouseUp, true);
     window.addEventListener('mouseover', this._onMouseOver, true);
+    this._selectInterval = setInterval(this._updateSelectedFiber, 200);
   };
 
   startProfiling = () => {
```
gaearon: This code is in `stopInspectingDOM`. GitHub is messing up the folding.

src/backend/agent.js
```diff
@@ -442,9 +450,20 @@ export default class Agent extends EventEmitter {
     event.stopPropagation();
 
     const target = ((event.target: any): HTMLElement);
-
     // Don't pass the name explicitly.
     // It will be inferred from DOM tag and Fiber owner.
     showOverlay(target);
+
+    const id = this.getIDForNode(target);
+    if (id !== null) {
```
bvaughn: Is the null check correct? If we aren't currently moused over an element that matches a fiber, do we still want to select the most recent one?

src/backend/agent.js
```diff
@@ -359,6 +363,11 @@ export default class Agent extends EventEmitter {
     window.removeEventListener('mousedown', this._onMouseDown, true);
     window.removeEventListener('mouseup', this._onMouseUp, true);
     window.removeEventListener('mouseover', this._onMouseOver, true);
+
+    this._updateSelectedFiber();
```
bvaughn: Is this right? (that we update selection after stop?)

bvaughn:
Btw I think I'm going to write a small throttle+memo util for #74. Maybe we could use it here too, since it seems conceptually similar.

src/backend/agent.js
gaearon: Do you think we should clear the selection completely?

src/backend/agent.js
gaearon: One last time since the interval is cancelled.

src/backend/agent.js
bvaughn: I don't think so. In this case, we should maybe just leave the most recently selected thing as-is?

src/backend/agent.js
bvaughn: I guess I was just thinking about the reasons `stopInspectingDOM` might be called:
1. The user clicked on a DOM node (in which case, it's already selected in the tree)
2. The user clicked on the the toggle in the DevTools

Seems like in either case, we wouldn't need to actually call `_updateSelectedFiber` anymore?

bvaughn: Ohh.... I'm wrong about the mouse down bit now I guess.

bvaughn:
Regarding my throttle comment above, maybe it would read a little better to do something like:

```js
import memoize from 'memoize-one';
import throttle from 'lodash.throttle';

class Agent {
  // ...

  _selectFiberForNode = throttle(
    memo((node: HTMLElement) => {
      const id = this.getIDForNode(node);
      if (id !== null) {
        this._bridge.send("selectFiber", id);
      }
    }),
    200
  );
}
```

Then `_onMouseDown` and `_onMouseOver` could both just call `this._selectFiberForNode(target)` directly.

gaearon:
Do you want to do this or I'll do it?

bvaughn:
Don't care, but I think it would be a  nice change to make before merging.

bvaughn:
I'll just do it.

PR #87, opened 2019-04-07
Bump overscan to 3

gaearon:
While this is subjective and depends on the laptop etc, I find this a bit less janky on fast scrolling. Rendering rows should be pretty cheap regardless (since we don't put much there) and I don't see regressions on large trees. WDYT?


bvaughn:
I don't feel strongly about this.

Issue #80, opened 2019-04-06
Selecting component at the bottom causes a jump


gaearon:
This only happens for components at the bottom of the scroll:

![Screen Recording 2019-04-06 at 04 27 pm](https://user-images.githubusercontent.com/810438/55671542-03430b00-5889-11e9-8d98-507bc717ac95.gif)

gaearon:
This seems to be caused by the List border — which messes up the calculation of offsets.

bvaughn:
Interesting! I didn't consider that aspect. I guess I can always move that border up a level

PR #88, opened 2019-04-07
Remove tree focus ring

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/80.

I figured this ring isn't really that useful because the first time you Tab into the tree, the first item gets selected. So you know it received focus. Chrome tree view doesn't have a focus ring either.

In the future, we might want to only enable tree arrow key navigation when the tree if focused. For example, that would allow us to repurpose up/down for owner stack navigation when owner stack is focused. If we were to do that, we could mark "focused" state of the tree with the default selection color, and "unfocused" state with a grey-ish color, similar to how Chrome Elements tab does. Even in that case, we wouldn't need a ring. But that's a future nice-to-have.

### DIFF
--- a/src/devtools/views/Components/Tree.css
+++ b/src/devtools/views/Components/Tree.css
@@ -24,10 +24,6 @@
   font-family: var(--font-family-monospace);
   font-size: var(--font-size-monospace-normal);
   line-height: var(--line-height-data);
-  border: 0.25rem solid transparent;
-}
-.List:focus-within {
-  border-color: var(--color-button-background-focus);
 }
 
 .InnerElementType:focus {


bvaughn:
> I figured this ring isn't really that useful because the first time you Tab into the tree, the first item gets selected. So you know it received focus.

I'm not sure I agree. I found it jarring when I tabbed into tree after an item was selected and there was no visual indication of where I was.

> Chrome tree view doesn't have a focus ring either.

Chrome has a separate focus background color for the selected item though. I guess we could add that to the theme.

gaearon:
Right — that's what I meant by

>If we were to do that, we could mark "focused" state of the tree with the default selection color, and "unfocused" state with a grey-ish color, similar to how Chrome Elements tab does. Even in that case, we wouldn't need a ring.

gaearon:
> I found it jarring when I tabbed into tree after an item was selected and there was no visual indication of where I was.

I agree it's not ideal btw. I find it less of an issue though because even when it's not directly focused, the arrows keep working. So at least it doesn't "lie" about whether it listens to the keyboard.

However, it's not idiomatic that it listens to up/down while it's not focused. And it even gets in the way for other fixes:

> For example, that would allow us to repurpose up/down for owner stack navigation when owner stack is focused. 

This is why I think "adding a grey unfocused row mode" and "only listening to arrow keys when focused" are parts of the same fix.

bvaughn:
Hm. What I was trying to say though is that I think the ring is important until/unless we differentiate between selected and selected+focused. Otherwise it's pretty jarring.

> I agree it's not ideal btw. I find it less of an issue though because even when it's not directly focused, the arrows keep working. So at least it doesn't "lie" about whether it listens to the keyboard.

That makes sense.

I think it would maybe be more intuitive if the arrows only work when you're focused on either the tree or the search input.

gaearon:
Maybe I wasn't very clear. Sorry. Here's how I'd rank different states in the order from unfortunate to desirable:

1. *Current:* Clicking sometimes shifts cursor (https://github.com/bvaughn/react-devtools-experimental/issues/80).

2. *This PR:* Tabbing first time gives you indication, but doesn't give you an indication next times. However, arrow keys always work so at least the indication doesn't "trick" you, and content is part of tab order.

3. *Possible future:* Focus changes current row color. Arrow keys only work in a tree it's focused. This would allow us to use arrow keys in more places — such as to navigating between owners.

gaearon:
Also I know this is super nitpicky but another thing that bugs me about the current ring is the pixel that appears between the border radius of the first item and the ring :P

Screen Shot 2019-04-07 at 18 57 33

I know this doesn't matter though.

bvaughn:
I guess I don't feel strongly enough to assert that one way is definitely
better or worse. Long term, no difference between focused and unfocused
feels objectively worse, but as a temporary step I don't oppose this change
I guess 😁

src/devtools/views/Components/Tree.css
```diff
@@ -24,10 +24,6 @@
   font-family: var(--font-family-monospace);
   font-size: var(--font-size-monospace-normal);
   line-height: var(--line-height-data);
-  border: 0.25rem solid transparent;
-}
-.List:focus-within {
-  border-color: var(--color-button-background-focus);
```
bvaughn: Just revert the border/padding change so we won't have the border offers l offset bug either?

gaearon: Maybe I misunderstood — but isn't this the fix for it? I thought I fixed it by removing borders (like in this PR). Or are you saying you want to add white paddings back? (They don't seem to influence that issue.)

src/devtools/views/Components/Tree.css
```diff
@@ -24,10 +24,6 @@
...
 }
```
bvaughn: Whoops I only saw the deleted border color line. Reviewing on mobile phone.

Issue #44, opened 2019-04-02
Convert inspected element (right panel) to use Suspense


bvaughn:
This could enable us to provide a better user experience than the flash of "Loading.." text when selection changes.

Ideally it would also be nice to...
* Globally throttle the frequency with which we request data (for any item). Currently we throttle by 1 second, but if you quickly select several items- the throttle is bypassed, so we may send multiple requests.
* Avoid re-requesting data for an item unless it has been modified (committed). This would require support from the renderer interface, but could fit in nicely with our Suspense invalidation strategy.
* Don't request props for elements hovered over while inspecting the DOM (using the selection tool). Only request them for elements that have been selected. We may also want to debounce the initial request to avoid sending a request for something a user quickly jumps over while keyboard navigating the tree. (This may be unnecessary if our throttling approach is solid enough.)

Maybe we could introduce the concept of a hovered element/index to the tree context. I'm not sure if this would make the already complicated code too complicated. Potentially it would... conceptually though, we could avoid requesting props for hovered elements entirely. We could even (potentially) style an unconfirmed hover _differently_ from the way we'd style a regular one. Suspense would be a prerequisite for this sort of feature, since we can't conditionally use hooks to e.g. subscribe.

bvaughn:
I'm going to pick this one up because I think it relates to #195 and I'm investigating that today.

bvaughn:
Here's a POC of the hover inspection
![hover-Kapture 2019-04-19 at 10 53 34](https://user-images.githubusercontent.com/29597/56436645-6cffe380-6291-11e9-8cb2-d6b25c463d16.gif)

bvaughn:
I think that avoiding sending inspection requests for "hovered" elements might actually end up being important for suspense too, so we don't end up suspending and re-suspending tons of times while using the inspection button.

PR #90, opened 2019-04-07
Move right pane loading state inside

gaearon:
This might be controversial but just throwing an idea out there.

I tried https://github.com/bvaughn/react-devtools-experimental/issues/44 in https://github.com/bvaughn/react-devtools-experimental/pull/79 but I'm still not quite sure how to make it work. However, there's a potentially simpler fix we could do that I think would improve the status quo.

In the status quo, *the whole layout* jumps on selection change. In particular, a section view gets replaced with a Loading text (which doesn't exist on the normal state in this place), and then immediately gets replaced back. That makes it jarring.

Screen Recording 2019-04-07 at 07 09 pm

I wonder if you would consider this an improvement:

Screen Recording 2019-04-07 at 07 11 pm

Here, the layout itself is more static. The `props` label is always guaranteed to be there — so I don't blow it away. It might be a bit weird to see `props` written alone but after a few minutes with it, I realized I prefer it a lot to the current behavior.

Of course, in longer term figuring out https://github.com/bvaughn/react-devtools-experimental/pull/79 would be nicer but this seems like an easy opportunity to improve this IMO.

### DIFF
--- a/src/devtools/views/Components/InspectedElementTree.js
+++ b/src/devtools/views/Components/InspectedElementTree.js
@@ -21,6 +21,7 @@ export default function InspectedElementTree({
   data,
   label,
   overrideValueFn,
+  isLoading = false,
   showWhenEmpty = false,
 }: Props) {
   const isEmpty = data === null || Object.keys(data).length === 0;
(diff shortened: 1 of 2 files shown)


src/devtools/views/Components/InspectedElementTree.js
```diff
@@ -37,14 +38,17 @@ export default function InspectedElementTree({
       <div className={styles.InspectedElementTree}>
         <div className={styles.HeaderRow}>
           <div className={styles.Header}>{label}</div>
-          {!isEmpty && (
+          <div style={{ visibility: isEmpty ? 'hidden' : '' }}>
```
gaearon: Without this, `props` label jumps between items that have props and items that don't have them because the "copy" button influences the section height.

bvaughn:
I'd kind of prefer we just migrate to suspense for this component rather
than iterating on its temporary loading state.

Issue #67, opened 2019-04-04
Unnecessary scroll up happens in some cases


gaearon:
Will try to get a better repro tomorrow. For now just writing up so I don’t forget. 

I noticed that if the height is small enough, sometimes messing with items close to the end of the list abruptly scrolls me just a bit above. In other words it doesn’t let me scroll to end myself.

bvaughn:
Could it be related to the currently selected item? The list scrolls to ensure that item is fully visible when the selection changes.

gaearon:
I don't recall changing the selection though.

gaearon:
Here's a gif. I'm not changing the selection, just scrolling.

![Screen Recording 2019-04-04 at 10 08 pm](https://user-images.githubusercontent.com/810438/55588670-42a51680-5726-11e9-8dd3-083449b7f87d.gif)

bvaughn:
Ah! Related cause then:
https://github.com/bvaughn/react-devtools-experimental/blob/6c5c2cd3ad6b8978414a8e8536778521352c804f/src/devtools/views/Components/Element.js#L46-L56

gaearon:
Yep. Looking at it, what happens is that

* Selected item goes slightly out of view.
* However, when we reach the bottom, for some reason it remounts. Maybe virtualization decides to render "last N items" where N is higher than normal count?
* Remounting causes the effect to fire again.

PR #91, opened 2019-04-07
Fix spurious autoscroll

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/67.

I'm not *super* happy with the fix because instead of overfiring, it will cause the autoscroll to underfire in some edge cases. (Such as searching for an item, scrolling away from it, and then searching for it again.) However that seems like a better tradeoff to me in the meantime than random jumps when scrolling.

We can't truly infer whether to auto-scroll or not from which virtual items are materialized, as we see from https://github.com/bvaughn/react-devtools-experimental/issues/67#issuecomment-480067969.

Another fix might be to set some flag [here](https://github.com/bvaughn/react-devtools-experimental/blob/a9c20f0b36366de81ab235cd8b3bd73fe454e513/src/devtools/views/Components/Tree.js#L44) and then reset it in the Element. However I'm not sure the order is deterministic enough — and handling all cases might turn it into a mess. Open to suggestions.

### DIFF
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -4,7 +4,7 @@ import React, {
   Fragment,
   useCallback,
   useContext,
-  useEffect,
+  useLayoutEffect,
   useMemo,
   useRef,
 } from 'react';
@@ -19,9 +19,11 @@ import styles from './Element.css';
 type Props = {
   index: number,
   style: Object,
+  // TODO: I can't get the correct type to work here:
+  data: Object,
 };
 
-export default function ElementView({ index, style }: Props) {
+export default function ElementView({ index, style, data }: Props) {
   const {
     baseDepth,
     getElementAtIndex,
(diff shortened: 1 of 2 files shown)


src/devtools/views/Components/Element.js
```diff
@@ -29,6 +30,7 @@ export default function ElementView({ index, style }: Props) {
     selectedElementID,
     selectElementByID,
   } = useContext(TreeContext);
+  const lastScrolledIDRef = useContext(ScrollAnchorContext);
```
bvaughn: Probably cheaper to just use `itemData` for this rather than adding a new context. Thoughts?

gaearon:
Pushed a fix that uses `itemData`. For some reason I couldn't get the prop to type correctly.

src/devtools/views/Components/Element.js
```diff
@@ -19,21 +19,23 @@ import styles from './Element.css';
 type Props = {
   index: number,
   style: Object,
+  // TODO: I can't get the correct type to work here:
+  data: Object,
```
bvaughn: Looks like the `itemData` generics got broken by a recent Flow upgrade :(

PR #92, opened 2019-04-08
Select DOM nodes on hover

gaearon:
I know you've considered and rejected this idea before, but I'm sending in case you might reconsider.

I've been watching @threepointone struggle to use new DevTools today. Initially, he couldn't find either the "inspect from DOM" button nor the "show in DOM" button because both are small icons that are easy to miss. He was sure neither of those are implemented yet.

Even when I pointed out the toolbar buttons, he was frustrated because every time he navigated in the tree, he had to press a button to see where he was. This made navigation itself difficult because *hovering nodes to see which one he wants to select* is part of his process.

This PR is a proposal to add behavior similar to what Chrome does in Elements tab, as well as what the old DevTools did:

1. When hovering a row, we highlight the corresponding DOM node.
2. We remove the highlighting when you mouseleave the tree.

There are a few smaller changes for this to work well. For example, I added a `scrollIntoView` option (`false` for hovers) and an `isSticky` option (`true` for hovers). Scrolling on hover is too disruptive. Automatic hiding is also disruptive in hover mode because it feels like the tooltip randomly disappears. However, I kept the existing "show in DOM" behavior. It's still useful if you want to scroll to an element.

## Inspecting Unfamiliar Trees: Before

Screen Recording 2019-04-08 at 02 33 PM

## Inspecting Unfamiliar Trees: After

Screen Recording 2019-04-08 at 02 32 PM

### DIFF
--- a/src/backend/views/Highlighter.js
+++ b/src/backend/views/Highlighter.js
@@ -18,7 +18,8 @@ export function hideOverlay() {
 
 export function showOverlay(
   element: HTMLElement | null,
-  componentName: string = ''
+  componentName: string | null,
+  isSticky: boolean
 ) {
   if (timeoutID !== null) {
     clearTimeout(timeoutID);
@@ -34,5 +35,7 @@ export function showOverlay(
 
   overlay.inspect(element, componentName);
 
-  timeoutID = setTimeout(hideOverlay, SHOW_DURATION);
+  if (!isSticky) {
+    timeoutID = setTimeout(hideOverlay, SHOW_DURATION);
+  }
 }
--- a/src/backend/views/Highlighter.js
+++ b/src/backend/views/Highlighter.js
@@ -19,7 +19,7 @@ export function hideOverlay() {
 export function showOverlay(
   element: HTMLElement | null,
   componentName: string | null,
-  isSticky: boolean
+  hideAfterTimeout: boolean
 ) {
   if (timeoutID !== null) {
     clearTimeout(timeoutID);
@@ -35,7 +35,7 @@ export function showOverlay(
 
   overlay.inspect(element, componentName);
 
-  if (!isSticky) {
+  if (hideAfterTimeout) {
     timeoutID = setTimeout(hideOverlay, SHOW_DURATION);
   }
 }
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -11,6 +11,7 @@ import React, {
 import { ElementTypeClass, ElementTypeFunction } from 'src/devtools/types';
 import { createRegExp } from '../utils';
 import { TreeContext } from './TreeContext';
+import { BridgeContext, StoreContext } from '../context';
 
 import type { Element } from './types';
 
@@ -31,6 +32,9 @@ export default function ElementView({ index, style, data }: Props) {
     selectedElementID,
     selectElementByID,
   } = useContext(TreeContext);
+  const bridge = useContext(BridgeContext);
+  const store = useContext(StoreContext);
+
   const element = getElementAtIndex(index);
 
   const id = element === null ? null : element.id;
@@ -82,6 +86,19 @@ export default function ElementView({ index, style, data }: Props) {
     [id, selectElementByID]
   );
 
+  const rendererID = store.getRendererIDForElement(element.id) || null;
+  const handleMouseEnter = useCallback(() => {
+    if (rendererID !== null) {
+      bridge.send('highlightElementInDOM', {
+        displayName: element.displayName,
+        id: element.id,
+        rendererID,
+        scrollIntoView: false,
+        isSticky: true,
+      });
+    }
+  }, [bridge, element, rendererID]);
+
   // Handle elements that are removed from the tree while an async render is in progress.
   if (element == null) {
     console.warn(`<ElementView> Could not find element at index ${index}`);
@@ -100,6 +117,7 @@ export default function ElementView({ index, style, data }: Props) {
   return (
     <div
       className={isSelected ? styles.SelectedElement : styles.Element}
+      onMouseEnter={handleMouseEnter}
       onMouseDown={handleMouseDown}
       onDoubleClick={handleDoubleClick}
       style={{
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -87,6 +87,8 @@ export default function ElementView({ index, style, data }: Props) {
   );
 
   const rendererID = store.getRendererIDForElement(element.id) || null;
+  // Individual elements don't have a corresponding leave handler.
+  // Instead, it's implemented on the tree level.
   const handleMouseEnter = useCallback(() => {
     if (rendererID !== null) {
       bridge.send('highlightElementInDOM', {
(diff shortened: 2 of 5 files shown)


src/devtools/views/Components/Element.js
```diff
@@ -66,6 +69,21 @@ export default function ElementView({ index, style }: Props) {
     [id, selectElementByID]
   );
 
+  const rendererID = store.getRendererIDForElement(element.id) || null;
```
bvaughn: nit `store.getRendererIDForElement` already returns `null` if it can't find a match

src/devtools/views/Components/Element.js
```diff
@@ -29,6 +30,8 @@ export default function ElementView({ index, style }: Props) {
     selectedElementID,
     selectElementByID,
   } = useContext(TreeContext);
+  const bridge = useContext(BridgeContext);
+  const store = useContext(StoreContext);
```
bvaughn: I wonder how much it will impact performance to have every list item reading from multiple contexts? Should we leverage the `itemData` prop for this?

The below diff has a Flow error that I would need to sort out, but as an example:
```diff
diff --git a/src/devtools/views/Components/Element.js b/src/devtools/views/Components/Element.js
index 6fd7db8..91c8064 100644
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -11,18 +11,20 @@ import React, {
 import { ElementTypeClass, ElementTypeFunction } from 'src/devtools/types';
 import { createRegExp } from '../utils';
 import { TreeContext } from './TreeContext';
-import { BridgeContext, StoreContext } from '../context';
 
 import type { Element } from './types';
+import type { ItemData } from './Tree';
 
 import styles from './Element.css';
 
 type Props = {
+  data: ItemData,
   index: number,
   style: Object,
 };
 
-export default function ElementView({ index, style }: Props) {
+export default function ElementView({ data, index, style }: Props) {
+  const { highlightElement } = data;
   const {
     baseDepth,
     getElementAtIndex,
@@ -30,8 +32,6 @@ export default function ElementView({ index, style }: Props) {
     selectedElementID,
     selectElementByID,
   } = useContext(TreeContext);
-  const bridge = useContext(BridgeContext);
-  const store = useContext(StoreContext);
 
   const element = getElementAtIndex(index);
 
@@ -60,6 +60,8 @@ export default function ElementView({ index, style }: Props) {
 
   // TODO Add click and key handlers for toggling element open/close state.
 
+  // Individual elements don't have a corresponding leave handler.
+  // Instead, it's implemented on the tree level.
   const handleMouseDown = useCallback(
     ({ metaKey }) => {
       if (id !== null) {
@@ -69,20 +71,12 @@ export default function ElementView({ index, style }: Props) {
     [id, selectElementByID]
   );
 
-  const rendererID = store.getRendererIDForElement(element.id) || null;
   // Individual elements don't have a corresponding leave handler.
   // Instead, it's implemented on the tree level.
-  const handleMouseEnter = useCallback(() => {
-    if (rendererID !== null) {
-      bridge.send('highlightElementInDOM', {
-        displayName: element.displayName,
-        id: element.id,
-        rendererID,
-        scrollIntoView: false,
-        isSticky: true,
-      });
-    }
-  }, [bridge, element, rendererID]);
+  const handleMouseEnter = useCallback(
+    () => highlightElement(((element: any): Element)),
+    [element, highlightElement]
+  );
 
   // Handle elements that are removed from the tree while an async render is in progress.
   if (element == null) {
diff --git a/src/devtools/views/Components/Tree.js b/src/devtools/views/Components/Tree.js
index ef6073e..af35102 100644
--- a/src/devtools/views/Components/Tree.js
+++ b/src/devtools/views/Components/Tree.js
@@ -12,14 +12,25 @@ import AutoSizer from 'react-virtualized-auto-sizer';
 import { FixedSizeList } from 'react-window';
 import { TreeContext } from './TreeContext';
 import { SettingsContext } from '../Settings/SettingsContext';
-import { BridgeContext } from '../context';
-import Element from './Element';
+import { BridgeContext, StoreContext } from '../context';
+import ElementView from './Element';
 import InspectHostNodesToggle from './InspectHostNodesToggle';
 import OwnersStack from './OwnersStack';
 import SearchInput from './SearchInput';
 
 import styles from './Tree.css';
 
+import type { Element } from './types';
+
+type HighlightElementFn = (element: Element) => void;
+
+export type ItemData = {|
+  baseDepth: number,
+  highlightElement: HighlightElementFn,
+  numElements: number,
+  getElementAtIndex: (index: number) => Element | null,
+|};
+
 type Props = {||};
 
 export default function Tree(props: Props) {
@@ -34,7 +45,8 @@ export default function Tree(props: Props) {
     selectPreviousElementInTree,
   } = useContext(TreeContext);
   const bridge = useContext(BridgeContext);
-  const listRef = useRef<FixedSizeList<any> | null>(null);
+  const store = useContext(StoreContext);
+  const listRef = useRef<FixedSizeList<ItemData> | null>(null);
   const treeRef = useRef<HTMLDivElement | null>(null);
 
   const { lineHeight } = useContext(SettingsContext);
@@ -93,15 +105,32 @@ export default function Tree(props: Props) {
     selectPreviousElementInTree,
   ]);
 
+  const highlightElement = useCallback<HighlightElementFn>(
+    (element: Element) => {
+      const rendererID = store.getRendererIDForElement(element.id);
+      if (rendererID !== null) {
+        bridge.send('highlightElementInDOM', {
+          displayName: element.displayName,
+          id: element.id,
+          rendererID,
+          scrollIntoView: false,
+          isSticky: true,
+        });
+      }
+    },
+    [bridge, store]
+  );
+
   // Let react-window know to re-render any time the underlying tree data changes.
   // This includes the owner context, since it controls a filtered view of the tree.
-  const itemData = useMemo(
+  const itemData = useMemo<ItemData>(
     () => ({
       baseDepth,
+      highlightElement,
       numElements,
       getElementAtIndex,
     }),
-    [baseDepth, numElements, getElementAtIndex]
+    [baseDepth, highlightElement, numElements, getElementAtIndex]
   );
 
   const handleMouseLeave = useCallback(() => {
@@ -128,7 +157,7 @@ export default function Tree(props: Props) {
               ref={listRef}
               width={width}
             >
-              {Element}
+              {ElementView}
             </FixedSizeList>
           )}
         </AutoSizer>
```

Maybe this isn't a big deal and we should just ignore it for now?

src/backend/views/Highlighter.js
```diff
@@ -18,7 +18,8 @@ export function hideOverlay() {
 
 export function showOverlay(
   element: HTMLElement | null,
-  componentName: string = ''
+  componentName: string | null,
+  isSticky: boolean
```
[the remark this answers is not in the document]
gaearon: How about `hideAfterTimeout` and invert the meaning?

src/devtools/views/Components/Element.js
gaearon: Copy pasted. Want me to change it in both places?

src/backend/views/Highlighter.js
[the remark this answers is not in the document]
bvaughn: 👍

src/devtools/views/Components/Element.js
bvaughn: Ah,  gotcha.  Yeah if you don't mind 😄

gaearon:
Addressed reviews except the context (as it's stable and should be free)

gaearon:
Rewrote with a separate button.

Issue #48, opened 2019-04-03
Preserve vertical scroll position when switching tabs


bvaughn:
Changing between the Elements and Profilers tab loses the vertical scroll position of each.

PR #93, opened 2019-04-08
Preserve state when switching tabs

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/48.

At first I considered only preserving the scroll state. However on a second thought I realized I'd prefer the tree to stay there. For example if I'm in a middle of inspecting some component's props and then jump to Profiler and back, I don't want my right pane scroll to reset either. As we add more interactivity to the right pane (e.g. expanding props / nested Hooks deeper), this problem will keep coming up.

So I just keep all tabs mounted once they are. I use the `hidden` prop which in Concurrent Mode should make inactive tab updates low pri. Seems like it doesn't break anything, but we can also switch to `display: none` if it ends up being a problem.

### DIFF
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -154,6 +154,14 @@ function createPanelIfReactLoaded() {
 
       initBridgeAndStore();
 
+      function ensureInitialHTMLIsCleared(container) {
+        if (container._hasInitialHTMLBeenCleared) {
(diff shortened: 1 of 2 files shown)


hristo-kanchev:
Nice! 
I was actually taking a look at this while on my lunch break.

Preserving the scroll state would've probably turned out messy.

PR #94, opened 2019-04-08
Rename "owner stack" label to "rendered by"

gaearon:
I think most people don't know what an "owner" is. Additionally, "stack" has unnecessary computer science connotations. Admittedly, there's a notion of "call stack" but especially beginners don't necessarily know that this is how you call the thing that you get after the error.

This wording seems a bit friendlier to me, and more immediately useful:

Screen Shot 2019-04-08 at 3 40 22 PM


bvaughn:
Seems reasonable

Issue #59, opened 2019-04-04
Keep DOM Elements in sync with Components selection


gaearon:
I miss the behavior of old DevTools that kept Chrome Elements and the React Components selection state in sync. I don't know how hard it would be to implement, but filing to look later.

Maybe somebody in open source can contribute?

bvaughn:
That's what this TODO is about btw:
https://github.com/bvaughn/react-devtools-experimental/blob/bfbb4ac545514d682a330087ff4be77aeab273f1/shells/browser/shared/src/main.js#L174

Avi98:
@bvaughn and @gaearon I would really love to contribute. If nobody is working on it, I can work on it.

gaearon:
Sure!

gaearon:
@Avi98 Did you get time to look into it yet? It seems pretty annoying to not have this feature so if you're busy and don't anticipate having time to build it, I'd like to get it in sooner.

Avi98:
Hi @gaearon, I've been busy lately but got some time to dig into it. I'll be needing some more time. Meanwhile, I don't want to be a blocker, So you can take up the work.

gaearon:
Did you manage to get something initial working? I'd like to finish this today but some pointers would be helpful.

Avi98:
No, I have not started anything, but I was going through react-devtools  code base to find out how they were doing it and found out this code `  chrome.devtools.panels.create('React', '', 'panel.html', function(panel) {
      var reactPanel = null;
      panel.onShown.addListener(function(window) {
        // when the user switches to the panel, check for an elements tab
        // selection
        window.panel.getNewSelection();
        reactPanel = window.panel;
        reactPanel.resumeTransfer();
      });
      panel.onHidden.addListener(function() {
        if (reactPanel) {
          reactPanel.hideHighlight();
          reactPanel.pauseTransfer();
        }
      });
    });`
 see if this could be of anyhelp

Avi98:
I have made a [gist](https://gist.github.com/Avi98/37cddcf7a486bf993d85f94c42391304)

gaearon:
Nice, thanks

gaearon:
https://github.com/bvaughn/react-devtools-experimental/pull/99

PR #99, opened 2019-04-08
Sync DevTools Elements and Components tabs

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/59.

This implements two-way syncing between the DevTools Elements tab and the React Components tab.

I think in most cases this is what you want. Especially now that we don't show the host nodes, and so you often need to switch to Elements to play with them.

Screen Recording 2019-04-08 at 08 47 PM

I'm not sure if there are any cases where the two-way binding gets in the way. Maybe we could offer an option to disable it as an escape hatch. Or maybe there's some smarter heuristic we could use.

### DIFF
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -85,6 +85,16 @@ function createPanelIfReactLoaded() {
           );
         });
 
+        // Remember if we should sync the browser DevTools to the React tab.
+        // We'll only do that if user intentionally chooses a different React component.
+        let lastSelectedID = null;
+        bridge.addListener('selectElement', ({ id }) => {
+          if (!hasReactSelectionChanged && lastSelectedID !== id) {
+            hasReactSelectionChanged = true;
+            lastSelectedID = id;
+          }
+        });
+
         // This flag lets us tip the Store off early that we expect to be profiling.
         // This avoids flashing a temporary "Profiling not supported" message in the Profiler tab,
         // after a user has clicked the "reload and profile" button.
@@ -180,12 +190,20 @@ function createPanelIfReactLoaded() {
         );
       }
 
+      let hasReactSelectionChanged = false;
+
       function maybeSetBrowserSelectionFromReact() {
+        // Don't change the browser element selection when navigating away
+        // from the Components tab if the user didn't change the React selection.
+        if (!hasReactSelectionChanged) {
+          return;
+        }
+        hasReactSelectionChanged = false;
         chrome.devtools.inspectedWindow.eval(
           '(window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$0 != null)' +
             '  ? (inspect(window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$0), true)' +
             '  : false',
-          (didChangeSelection, error) => {
+          (_, error) => {
             if (error) {
               console.error(error);
             }
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -162,10 +162,43 @@ function createPanelIfReactLoaded() {
         container._hasInitialHTMLBeenCleared = true;
       }
 
+      function maybeSetReactSelectionFromBrowser() {
+        // When the user chooses a different node in the browser Elements tab,
+        // copy it over to the hook object so that we can sync the selection.
+        chrome.devtools.inspectedWindow.eval(
+          // Don't reset selection if it didn't change in Elements tab.
+          '(window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$0 !== $0)' +
+            '  ? (window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$0 = $0, true)' +
+            '  : false',
+          (didChangeSelection, error) => {
+            if (error) {
+              console.error(error);
+            } else if (didChangeSelection) {
+              bridge.send('syncSelectionFromBrowserTools');
+            }
+          }
+        );
+      }
+
+      function maybeSetBrowserSelectionFromReact() {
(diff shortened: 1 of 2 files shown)


gaearon:
One plausible heuristic could be “don’t change DOM child if we manually selected a different DOM node and didn’t change the component selection since last time”.

Then at least it won’t change if you intentionally choose a different one. 

I’m too tired to implement it today but it would likely be a good change.

gaearon:
OK, I pushed up a version that tracks whether React selection changed before trying to reset the browser one. I think this should be good enough now.

shells/browser/shared/src/main.js
```diff
@@ -85,6 +85,16 @@ function createPanelIfReactLoaded() {
           );
         });
 
+        // Remember if we should sync the browser DevTools to the React tab.
+        // We'll only do that if user intentionally chooses a different React component.
+        let lastSelectedID = null;
+        bridge.addListener('selectElement', ({ id }) => {
+          if (!hasReactSelectionChanged && lastSelectedID !== id) {
```
bvaughn: This looks wrong. We won't ever update `lastSelectedID` after the first time?

bvaughn: Ah. I see what you're doing. I think the name of `lastSelectedID` just  confused me. It seemed like we should be updating it each time a new "selectElement" message was sent, but in this case we're just using it to decide when to set `hasReactSelectionChanged`.

I don't think we need it. Pretty sure we can just set `hasReactSelectionChanged = true` any time a "_selectElement_" message is sent. (We only send this message when a _new_ element is selected in the tree.)

```js
bridge.addListener('selectElement', ({ id }) => {
  hasReactSelectionChanged = true;
});
```

src/backend/agent.js
```diff
@@ -87,6 +87,10 @@ export default class Agent extends EventEmitter {
     bridge.addListener('startProfiling', this.startProfiling);
     bridge.addListener('stopInspectingDOM', this.stopInspectingDOM);
     bridge.addListener('stopProfiling', this.stopProfiling);
+    bridge.addListener(
+      'syncSelectionFromBrowserTools',
+      this.syncSelectionFromBrowserTools
```
bvaughn: trivial nit: How do you feel about replacing "_BrowserTools_" with "_ElementsPanel_" ? Seems more specific.

shells/browser/shared/src/main.js
gaearon: That was my first attempt and it didn’t work because this message would get sent in other cases (such as when we sync DevTools *to* match the DOM node).

shells/browser/shared/src/main.js
```diff
@@ -162,10 +162,43 @@ function createPanelIfReactLoaded() {
...
+            }
+          }
+        );
+      }
+
       let currentPanel = null;
 
       chrome.devtools.panels.create('⚛ Components', '', 'panel.html', panel => {
         panel.onShown.addListener(panel => {
+          maybeSetReactSelectionFromBrowser();
```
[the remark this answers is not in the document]
gaearon: Do you mean you’d like to change something? Not sure I understood.

src/backend/agent.js
gaearon: Wanted to make it somehow clear that this isn’t taking about any of our own panels.

shells/browser/shared/src/main.js
bvaughn: Hmm...it seemed to work in my testing.

src/backend/agent.js
```diff
@@ -292,6 +296,14 @@ export default class Agent extends EventEmitter {
     if (renderer == null) {
       console.warn(`Invalid renderer id "${rendererID}" for element "${id}"`);
     } else {
+      // Update the active DOM node on the global hook object.
+      // The content script will read this to update window.$0
+      // when we switch tabs.
+      let node: HTMLElement | null = null;
+      node = ((renderer.findNativeByFiberID(id): any): HTMLElement);
```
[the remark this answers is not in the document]
gaearon: Bad copy paste

shells/browser/shared/src/main.js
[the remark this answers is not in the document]
bvaughn: No. I was just saying that I like how this is implemented, only updating once rather than on-selection.

src/backend/agent.js
bvaughn: Yeah, I guess "browser tools" just means nothing to me.

shells/browser/shared/src/main.js
gaearon: It’s a bit tricky to test this specific piece. You’ll want to:

* select DOM node
* switch to Components
* switch back
* select a node that’s unrelated but doesn’t belong to another React component (like document head or some host sibling that’s still in the same React component)
* switch again to Components
* switch back again

The intended result is that DOM selection should stay on the host node you last chose. But with the approach you’re suggesting it would get reset again.

src/backend/agent.js
gaearon: Maybe “native elements tab”?

bvaughn: 👍

shells/browser/shared/src/main.js
gaearon: Hmm. Actually I can't repro this anymore. Maybe I did something wrong when I tested. Your suggestion sounds good.

gaearon:
Here's a quirky scenario:

* Select DOM node that belongs to a component but isn't its main node
* Switch to Components
* Switch back to Elements

I think in this case we shouldn't reset the selection to component's main node. But currently we do.

In other words, I wouldn't expect just switching between here, there, and back, to ever change the selection.

gaearon:
With last change I feel good about this

Issue #102, opened 2019-04-08
Can't switch from Components tab to Console tab


gaearon:
I broke this with https://github.com/bvaughn/react-devtools-experimental/pull/99.

Calling `inspect` forces Chrome to always switch to Elements tab. As a result navigation to other tabs are broken. We either need to find a way to detect switching *to Elements tab specifically* or, sadly, disable the syncing from React to DOM.

PR #103, opened 2019-04-09
Make DOM selection binding one-way

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/102 by making syncing one-way.

I'll talk to DevTools folks in case there's some way to restore the two-way binding later. Ideally I want a version of `inspect(foo)` that doesn't force a tab switch. Maybe they'll add it in the future.

The upside is apparently there's `chrome.devtools.panels.elements.onSelectionChanged` API that makes the other half easier. So I switched to that here too.

### DIFF
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -85,12 +85,6 @@ function createPanelIfReactLoaded() {
           );
         });
 
-        // Remember if we should sync the browser DevTools to the React tab.
-        // We'll only do that if user intentionally chooses a different React component.
-        bridge.addListener('selectElement', () => {
-          hasReactSelectionChanged = true;
-        });
-
         // This flag lets us tip the Store off early that we expect to be profiling.
         // This avoids flashing a temporary "Profiling not supported" message in the Profiler tab,
         // after a user has clicked the "reload and profile" button.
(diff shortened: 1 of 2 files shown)


gaearon:
I'll get this in because it's pretty much unusable otherwise. :-(

Happy to follow up on review comments. I'll also follow up with Chrome to file a feature request for a version of `inspect` that doesn't switch tabs.

shells/browser/shared/src/main.js
```diff
@@ -168,51 +162,39 @@ function createPanelIfReactLoaded() {
...
-          (_, error) => {
-            if (error) {
-              console.error(error);
-            }
-          }
-        );
+        isListeningToNativeSelectionChange = true;
+        setReactSelectionFromBrowser();
+        chrome.devtools.panels.elements.onSelectionChanged.addListener(() => {
+          setReactSelectionFromBrowser();
```
bvaughn: Hm... I understand why we had to remove the "two-way binding" but now we're updating the selection every time it changes in the Elements panel, rather than just when the React tab is about to be shown again? Why?

gaearon: Oh, I see, you want to avoid calling the bridge? I'll fix that.

PR #107, opened 2019-04-09
pointer-events: none to fix flicker on Overlay "tip"

sophiebits:
If click-to-inspect then hover an element in just the right place, it flickers between two elements because this "tip" element catches the hover. This should fix it.


sophiebits:
Not my fault that it flickers when I hover over stuff. (:

PR #108, opened 2019-04-09
Use @reach MenuButton for owner stack menu [WIP]

bvaughn:
I find myself implementing arrow key navigation in several places. It would be nice to use a reusable controller for this, and the [Reach UI `MenuButton`](https://ui.reach.tech/menu-button/) looks pretty promising for this.

Couple of blockers:
- [x] Reach has some pretty heavy-handed default styles. I need to think a bit more about the best way to override them to match our themes.
- [ ] The `Portal` component currently appends content to the root `document` rather than the owner, which breaks the extension. I've submitted a PR to Reach to fix this. (https://github.com/reach/reach-ui/pull/137)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,6 +3,7 @@ node_modules
 shells/browser/chrome/build
 shells/browser/firefox/build
 shells/dev/build
+vendor
 
 package-lock.json
 yarn.lock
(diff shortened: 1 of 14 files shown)


bvaughn:
I'm going to go ahead and fork `@reach/portal` (locally) for now to unblock this PR.

PR #109, opened 2019-04-09
Add collapsible icon for object and array types

lucasecdb:
I've got it working, but the arrow is kinda small on comfortable mode, maybe I shouldn't be using the `ButtonIcon` here?

inspect-collapse

This is comfortable mode:

image

### DIFF
--- a/src/devtools/views/Components/KeyValue.css
+++ b/src/devtools/views/Components/KeyValue.css
@@ -26,3 +26,13 @@
   color: var(--color-dimmer);
   font-style: italic;
 }
+
+.Opener {
+  padding: 0.125rem;
+  margin-left: -1rem;
+}
+
+.Opener svg {
+  width: 0.75rem;
+  height: 0.75rem;
+}
--- a/src/devtools/views/Components/KeyValue.css
+++ b/src/devtools/views/Components/KeyValue.css
@@ -32,5 +32,6 @@
   display: inline-flex;
   width: 1rem;
   height: 1rem;
+  padding: 0;
   color: var(--color-expand-collapse-toggle);
 }
--- a/src/devtools/views/Components/KeyValue.css
+++ b/src/devtools/views/Components/KeyValue.css
@@ -1,5 +1,6 @@
 .Item {
   display: flex;
+  --color-expand-collapse-toggle: var(--color-dim);
 }
 
 .Name {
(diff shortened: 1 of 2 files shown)


src/devtools/views/Components/KeyValue.css
```diff
@@ -26,3 +26,13 @@
   color: var(--color-dimmer);
   font-style: italic;
 }
+
+.Opener {
+  padding: 0.125rem;
+  margin-left: -1rem;
+}
+
+.Opener svg {
```
lucasecdb: I don't know if this is desirable, since we are using css modules, but I think maybe the only other way is to allow `ButtonIcon` to receive a className?

bvaughn: I'd rather avoid descendant selectors since they're slow. The pattern we follow in other components (e.g. `Button`, `Icon`) is to accept an optional `className` prop. `ButtonIcon` just didn't have a need for one before.

bvaughn:
I think the most common UI for this sort of toggle is an arrow pointing right when collapsed and down when expanded. (Rather than up and down.)

I added "_collapsed_" and "_expanded_" icons with PR #116 icons that you could use for these 😄

src/devtools/views/Components/KeyValue.css
```diff
@@ -26,3 +26,13 @@
   color: var(--color-dimmer);
   font-style: italic;
 }
+
+.Opener {
+  padding: 0.125rem;
+  margin-left: -1rem;
```
bvaughn: Strong preference to avoid negative margins.

lucasecdb:
@bvaughn oh, nice! I'll update the icons

src/devtools/views/Components/KeyValue.css
lucasecdb: just out of curiosity, why is that?

bvaughn: When things overflow their container, it makes it harder to reason about the layout. Things can also get clipped pretty easily if any ancestors hide overflow.

lucasecdb:
there's something off in the hooks inspector, I'll investigate later today

image

bvaughn:
Let me take a pass at it first. (I'm already mid review.)

lucasecdb:
I've found that changing the following line

https://github.com/bvaughn/react-devtools-experimental/blob/0aacc2547edb5252c3987a716e7852a12903ed98/src/devtools/views/Components/HooksTree.js#L169

to `depth={1}` solves the issue in the state hook, but I don't know if the 0 depth is intencional.

bvaughn:
This is really nice. Thanks so much for this PR!

bvaughn:
We could take collapsing a step further inside of hooks if folks want to. At the moment, I just made sure things aligned properly on the left, but we don't collapse by default inside of custom hooks.

bvaughn:
I might take a quick pass at that ^

bvaughn:
Actually no I'm going to just file an issue for it: #127

PR #113, opened 2019-04-09
Add inactive color on components when window isn't focused

### LINKED ISSUE #106: Hard to tell if devtools is focused
In old devtools, the focused row becomes gray if your focus is not on the devtools:

image

And blue when it is.

In new devtools, you can't tell if you're focused. This is frustrating. For example:

1. Click the inspect crosshairs button
2. Click an element in the page
3. Press Up or Down arrows to move the selection. It fails. (You think your selection is on the devtools but it's not!)

(Bonus feature request: If it is possible to focus the panel automatically after click-to-inspect, that would be awesome too.)

lucasecdb:
fixes #106 

not super happy with the colors I've chosen, but here's what they look like

image

image

### DIFF
--- a/src/devtools/views/Components/Element.css
+++ b/src/devtools/views/Components/Element.css
@@ -1,5 +1,6 @@
 .Element,
-.SelectedElement {
+.SelectedElement,
+.InactiveElement {
   border-radius: 0.25em;
   white-space: nowrap;
   line-height: var(--line-height-data);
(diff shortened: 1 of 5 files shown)


gaearon:
It shouldn't just be document, but the actual tree. If you focus into the right pane it should become grey. See how Chrome tools work.

lucasecdb:
oh, okay! makes sense

bvaughn:
Thanks! 😄

bvaughn:
Really nice to have the active/inactive color distinction in the tree now. Thanks again!

PR #114, opened 2019-04-09
Send the bridge sync event lazily

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/pull/103#discussion_r273619654.

### DIFF
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -166,35 +166,35 @@ function createPanelIfReactLoaded() {
         // When the user chooses a different node in the browser Elements tab,
         // copy it over to the hook object so that we can sync the selection.
         chrome.devtools.inspectedWindow.eval(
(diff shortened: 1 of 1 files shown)


shells/browser/shared/src/main.js
```diff
@@ -166,35 +166,35 @@ function createPanelIfReactLoaded() {
...
-      // select the corresponding React component.
-      let isListeningToNativeSelectionChange = false;
-      function ensureListeningToNativeSelectionChange() {
-        if (isListeningToNativeSelectionChange) {
-          return;
-        }
-        isListeningToNativeSelectionChange = true;
+      setReactSelectionFromBrowser();
+      chrome.devtools.panels.elements.onSelectionChanged.addListener(() => {
         setReactSelectionFromBrowser();
```
gaearon: I keep executing this eagerly because setting a field is super cheap and lets us have less of a delay when we actually do change the tab. The bridge event gets delayed until we show the tab tho.

Issue #115, opened 2019-04-09
Add test coverage for render and store


bvaughn:
There are several parts of the code that have gotten quite complex: tree serialization logic, "flatting" of the tree to a list for windowing, suspense fallback state, etc.

It would be nice if we had some tests that accepted a React tree, ran it through the renderer and into the store, and then asserted the output (via snapshots). The tests could also cover expanding/collapsing nodes and how this affected the tree, search, etc.

bvaughn:
#258 has some related work

bvaughn:
I think this is good enough for now.

PR #101, opened 2019-04-08
Pressing next forces search to select

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/100.

This becomes more important in light of merging https://github.com/bvaughn/react-devtools-experimental/pull/99 which might cause selection elsewhere in the tree. You want to be able to get back to your search results.

Screen Recording 2019-04-09 at 12 05 am

Issue #78, opened 2019-04-06
Consider adjusting scroll to preserve position when there was no user action


gaearon:
Noticed this debugging DevTools itself.

While not critical, I think it would be nice if changes that were not initiated by the developer (such as actual scrolling or clicks) wouldn't cause the selected element to jump. In other words, I would expect that below the first scroll "page", prepends anywhere earlier in the tree would keep the perceived scroll position instead of shifting it up and down.

Not sure if it's inherent to the current virtual scrolling implementation or something that would be easy to fix. Maybe virtual scrolling actually makes it easier to fix because we control everything?

![Screen Recording 2019-04-06 at 03 44 pm](https://user-images.githubusercontent.com/810438/55671010-04713980-5883-11e9-825d-f098dd620e1d.gif)

bvaughn:
My initial/gut reaction is that scroll anchoring seems kind of tricky to get right. Not sure what you're showing above (using a windowed DevTools to inspect a windowed DevTools) is all that representative of a common use case.

gaearon:
My point wasn't to show windowed DevTools, it's just the only example I have on the non-work computer of something in a DEV mode. I don't see how "items being appended and removed above my current selection" is an uncommon case. Unless I'm missing something, this would happen in any React application all the time, depending on what I select. 

I've had similar issues debugging FB where components changing above in the tree would make the selection jump. My concern is it's disorienting. I'm not saying it's a mustfix.

If you prefer, I can keep a separate record of "not must fix" issues somewhere. I thought that it's useful to list all UX annoyances — but if this is stressful I can close and stop filing those.

bvaughn:
> If you prefer, I can keep a separate record of "not must fix" issues somewhere. I thought that it's useful to list all UX annoyances — but if this is stressful I can close and stop filing those.

We chatted about this out of band but in case anyone sees this interaction on GH and  wonders:

I think it's good that you're filing issues to discuss ideas you have like this. 👍 

I also hope it's okay for me to respond with my honest feedback when I read them. (And hopefully it won't sound like I'm put out or being dismissive or anything. If it ever does- I apologize. That's not my intent.)

I sometimes have wary initial reactions to certain UX changes because I imagine they'll add a lot of complexity or require a lot of effort to implement, but this initial assessment isn't always correct. Sometimes it just means that I haven't considered the problem long enough or from the right angle.

gaearon:
[...]

PR #116, opened 2019-04-10
Add collapse/toggle UI to Tree

### LINKED ISSUE #105: Trees aren't collapsible?
Not sure if I'm missing something obvious, but I don't see any disclosure triangles or similar ways to expand and collapse parts of the tree. This means that I always see every component at every level. Is that expected?

bvaughn:
Resolves #105 (I think?)

Checklist:
- [x] Add toggle UI
- [x] Update `Store` elements count and `getElementAtIndex` to account  for collapsed nodes
- [x] Update `Store` tree operations to account for collapsed state
- [x] Update `TreeContext` search logic to account for collapsed nodes
- [x] Make left/right arrows toggle collapsed state as well

### DIFF
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -71,10 +71,6 @@ export default class Store extends EventEmitter {
   // When profiling is in progress, operations are stored so that we can later reconstruct past commit trees.
   _isProfiling: boolean = false;
 
-  // Total number of visible elements (within all roots).
-  // Used for windowing purposes.
-  _numElements: number = 0;
-
   // Suspense cache for reading profilign data.
   _profilingCache: ProfilingCache;
 
@@ -111,6 +107,10 @@ export default class Store extends EventEmitter {
   _supportsFileDownloads: boolean = false;
   _supportsProfiling: boolean = false;
   _supportsReloadAndProfile: boolean = false;
+
+  // Total number of visible elements (within all roots).
+  // Used for windowing purposes.
+  _weightAcrossRoots: number = 0;
 
   constructor(bridge: Bridge, config?: Config) {
     super();
@@ -198,7 +198,7 @@ export default class Store extends EventEmitter {
   }
 
   get numElements(): number {
-    return this._numElements;
+    return this._weightAcrossRoots;
   }
 
   get profilingCache(): ProfilingCache {
@@ -429,7 +429,7 @@ export default class Store extends EventEmitter {
 
       const weightDelta = isCollapsed ? 1 - element.weight : element.weight - 1;
 
-      this._numElements += weightDelta;
+      this._weightAcrossRoots += weightDelta;
 
       let parentElement = this._idToElement.get(element.parentID);
       while (parentElement != null) {
@@ -760,7 +760,7 @@ export default class Store extends EventEmitter {
 
       // Additions and deletions within a collapsed subtree should not affect the overall number of elements.
       if (!isInsideCollapsedSubTree) {
-        this._numElements += weightDelta;
+        this._weightAcrossRoots += weightDelta;
       }
     }
 
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -430,9 +430,10 @@ export default class Store extends EventEmitter {
   toggleIsCollapsed(id: number, isCollapsed: boolean): void {
     const element = this.getElementByID(id);
     if (element !== null) {
+      const oldWeight = element.isCollapsed ? 1 : element.weight;
       element.isCollapsed = isCollapsed;
-
-      const weightDelta = isCollapsed ? 1 - element.weight : element.weight - 1;
+      const newWeight = element.isCollapsed ? 1 : element.weight;
+      const weightDelta = newWeight - oldWeight;
 
       this._weightAcrossRoots += weightDelta;
 
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -353,12 +353,7 @@ export default class Store extends EventEmitter {
           break;
         }
         const child = ((this._idToElement.get(childID): any): Element);
-
-        // We intentionally ignore collapsed state when determining an item's index.
-        // We only do this for the direct path to an element.
-        // That's because the index of the element is meaningless if it's inside of a collapsed tree.
-        // If this index is used to display the element, the caller should also un-collapse its ancestors.
-        index += child.weight;
+        index += child.isCollapsed ? 1 : child.weight;
       }
 
       previousID = current.id;
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -353,7 +353,7 @@ export default class Store extends EventEmitter {
           break;
         }
         const child = ((this._idToElement.get(childID): any): Element);
-        index += child.weight;
+        index += child.isCollapsed ? 1 : child.weight;
       }
 
       previousID = current.id;
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -287,16 +287,18 @@ export default class Store extends EventEmitter {
     let currentElement = ((this._idToElement.get(firstChildID): any): Element);
     let currentWeight = rootWeight;
     while (index !== currentWeight) {
-      for (let i = 0; i < currentElement.children.length; i++) {
+      const numChildren = currentElement.children.length;
+      for (let i = 0; i < numChildren; i++) {
         const childID = currentElement.children[i];
         const child = ((this._idToElement.get(childID): any): Element);
-        const { weight } = child;
-        if (index <= currentWeight + weight) {
+        const childWeight = child.isCollapsed ? 1 : child.weight;
+
+        if (index <= currentWeight + childWeight) {
           currentWeight++;
           currentElement = child;
           break;
         } else {
-          currentWeight += weight;
+          currentWeight += childWeight;
         }
       }
     }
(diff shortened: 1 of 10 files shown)


src/devtools/views/Components/TreeContext.js
```diff
@@ -298,10 +299,25 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
             if (prevSearchIndex === null) {
               searchIndex = 0;
             } else {
-              searchIndex = Math.min(
-                ((prevSearchIndex: any): number),
-                searchResults.length - 1
-              );
+              // Changes in search index or typing should override the selected element.
+              // The one exception is when a search is broadened (e.g. "Cat" -> "Ca").
```
sophiebits: You might be editing anywhere in a search query, not just the end. To me, it would seem a slightly better heuristic would be:

When the search query changes…
1. If the currently selected element is one of the search results, keep the selection unchanged.
2. Otherwise, move the selection to the first result.

src/devtools/store.js
```diff
@@ -420,6 +422,28 @@ export default class Store extends EventEmitter {
     this.emit('isProfiling');
   }
 
+  toggleIsCollapsed(id: number, isCollapsed: boolean): void {
+    const element = this.getElementByID(id);
+    if (element !== null) {
+      element.isCollapsed = isCollapsed;
+
+      const weightDelta = isCollapsed ? 1 - element.weight : element.weight - 1;
```
sophiebits: This might be more obviously correct:

```js
const oldWeight = element.isCollapsed ? 1 : element.weight;
element.isCollapsed = isCollapsed;
const newWeight = element.isCollapsed ? 1 : element.weight;
const weightDelta = newWeight - oldWeight;
```

src/devtools/store.js
```diff
@@ -420,6 +422,28 @@ export default class Store extends EventEmitter {
...
+
+      this._numElements += weightDelta;
```
sophiebits: This logic might be a little easier to understand if you instead make a synthetic id-0 root node that has all the React roots as its children? Then you also won't need the special case in onBridgeOperations.

(This also seems good though. It wasn't obvious to me the purpose of this property though – renaming to `_weightAcrossRoots` or similar might be easier to understand.)

src/devtools/store.js
sophiebits: Thinking more on this, I don't love that by doing delta updates, this allows for errors to compound if any of the logic is wrong.

I think my preference would be to re-derive it each time. One easy way to do this is to have it be a lazy getters (or equivalent):

```js
function getVisibleWeight(el) {
  if (el.cachedVisibleWeight) {
    return el.cachedVisibleWeight;
  }
  return el.cachedVisibleWeight = el.isCollapsed ? 1 : 1 + sum(getVisibleWeight() of children);
}
```

When you change the children of any node or expand/collapse, you invalidate the cached weights of all the ancestors. Notably:
1. You do *not* need to recompute them all eagerly, and it is better not to. That way, if you change two parts of a tree, you only recompute the ancestor once.
2. When you are traversing ancestors, you can stop once you hit an ancestor that is already invalidated. That means you don't need to traverse all the way to the root for every change in a deep tree, only the first.

bvaughn: > Thinking more on this, I don't love that by doing delta updates, this allows for errors to compound if any of the logic is wrong.

AFK and so I'm not really responding to the review as a whole, just cherry picking this specific detail of it. In general, if any of the "logic is wrong", the tree is screwed. Everything is tuned for optimal performance (in some cases at the slight cost of readability) as best I know how.

I plan to start adding tests soon to cover the various types of tree mutations to reduce the likelihood of a regression

sophiebits:
Bug: If you click-to-inspect, it doesn't take into account collapsed nodes. Repro:

1. Open https://react-devtools-experimental.now.sh/
2. Collapse `<List>`
3. Click-to-inspect "First" in the page

Actual behavior: `<InteractionTracing>` gets selected
Ideal behavior: ListItem 1 is selected (and its ancestors expanded)

src/devtools/store.js
sophiebits: > Everything is tuned for optimal performance (in some cases at the slight cost of readability) as best I know how.

Well, you're using React so I assume you're taking at least _some_ preference for readability/maintainability/verifiable-correctness over the slightly faster possible speed that using solely imperative DOM mutations would provide.

I'm sympathetic to wanting it to be fast which is why I mentioned perf, but when it is possible to write code in a simpler way that is easier to verify the correctness of, I think that's better unless it is an extreme perf hotspot (which I really wouldn't expect for this case).

sophiebits:
I guess it is due (in part) to getIndexOfElementID not taking collapsed-ness into account.

src/devtools/views/Components/TreeContext.js
gaearon: Ah good point. I haven’t thought about this when I added it.

gaearon:
Should it start out collapsed? That would mitigate https://github.com/bvaughn/react-devtools-experimental/issues/78.

gaearon:
Bug: after click-to-inspect, arrow keys stop working.

gaearon:
Bug: clicking "right arrow" selects next sibling if there's no children. But clicking "left arrow" only goes back to parents. As a result, clicking "left" doesn't undo clicking "right" which is surprising. If we implement toggling, I think it's a good idea to follow the Chrome key behavior.

gaearon:
Bug:

1. Click ObjectProps collapse button
2. Click Contexts

Expected: Context gets selected
Actual: LegacyContextProvider gets selected

Similarly:

1. Click Contexts collapse button
2. Click CustomHooks

Expected: CustomHooks gets selected
Actual: it jumps to FuncitonComponent

src/devtools/views/Components/TreeContext.js
gaearon: Fixing this in https://github.com/bvaughn/react-devtools-experimental/pull/101.

bvaughn: #101 should stack on top of this PR, right?

gaearon: Not sure what you mean. I sent #101 a while ago — it's an independent fix. I just added a new commit to it that helps this particular heuristic. I guess you could do it in either order but I think merging #101 doesn't need to wait for this PR.

src/devtools/store.js
bvaughn: Hmm. We may have to agree to disagree on the right trade off between readability and performance in this case. In the case of the `Store` and op-codes, I intentionally chose a _less readable_ approach so that I could optimize perf as much as possible. (We make a lot of similar trade offs in the React source code.)

Amortizing the cost of these tree mutations is a core part of the design, so that interactions with the tree (especially scrolling and windowing related logic) can be as fast as possible.

src/devtools/store.js
bvaughn: Sure. I don't mind renaming it. There's an inline comment that explains what it is too.

> Total number of visible elements (within all roots).
> Used for windowing purposes.

bvaughn:
> Should it start out collapsed? That would mitigate #78.

I thought about this briefly, but I think it's kind of annoying to have to expand to see anything useful. I think I prefer the expanded by default approach. :smile:

> Bug: after click-to-inspect, arrow keys stop working.

This bug exists in master. Let's follow up outside of this PR.

src/devtools/views/Components/TreeContext.js
bvaughn: My bad Dan. I was confused. I overlooked #101 somehow and thought it was a PR you created overnight. 😄  I just looked at it and it looks good, so I'll rebase this on top of it.

bvaughn:
I'm going to dig into [this bug](https://github.com/bvaughn/react-devtools-experimental/pull/116#issuecomment-481518806) and [this bug](https://github.com/bvaughn/react-devtools-experimental/pull/116#issuecomment-481687390) now. Thanks for spotting them! 😄

src/devtools/store.js
sophiebits-humu: My suggestion also amortizes the calculations. :)

Fine to stick with your way, but I stand by my suggestion in the first comment then.

bvaughn: Thats fair 👍 I don't mind making that change.

bvaughn:
I think the bugs reported above (not counting the two that exist in master) have now been fixed.

src/devtools/store.js
```diff
@@ -351,6 +353,11 @@ export default class Store extends EventEmitter {
           break;
         }
         const child = ((this._idToElement.get(childID): any): Element);
+
+        // We intentionally ignore collapsed state when determining an item's index.
+        // We only do this for the direct path to an element.
+        // That's because the index of the element is meaningless if it's inside of a collapsed tree.
+        // If this index is used to display the element, the caller should also un-collapse its ancestors.
         index += child.weight;
```
sophiebits: Wait, doesn’t this need to honor collapsed because it counts previous siblings of ancestors, not just ancestors?

The root weight calculation later in the function too.

bvaughn: > Wait, doesn’t this need to honor collapsed because it counts previous siblings of ancestors, not just ancestors?

Hmm...yeah, that was my initial intuition as well, but then I convinced myself it wasn't true. I think you're right though that something might be off here. Lemme look at it a bit.

> The root weight calculation later in the function too.

Hmm... roots themselves are invisible. They will never be collapsed.

bvaughn: Yeah I think I may have done the _exact opposite_ thing here. 😄 I really need to write some good tests for this (#115)

bvaughn: Okay. I reverted that misguided bit. Nice catch 😄

bvaughn:
Sweeeeeeet!

I really appreciate your and Dan's feedback and bug reports ❤️  It's nice working with you again on something.

Issue #121, opened 2019-04-10
Improve navigation with left/right arrows


gaearon:
Spill-over from https://github.com/bvaughn/react-devtools-experimental/issues/105.

- [ ] Pressing left/right should always collapse/expand and navigate to parent/child next. It shouldn't navigate to siblings as it currently does. (https://github.com/bvaughn/react-devtools-experimental/pull/116#issuecomment-481681711)
- [x] After click-to-inspect, arrow keys stop working altogether.

bvaughn:
> After click-to-inspect, arrow keys stop working altogether.

For what it's worth, I'm pretty sure this is just a focus issue. Our key down handler in the `Tree` ignores events coming from an `<input>` (since that would be annoying in most cases).  The click-to-inspect button is a toggle (checkbox), so it causes subsequent key down events to be ignored until something else is focused.

bvaughn:
>  Pressing left/right should always collapse/expand and navigate to parent/child next. It shouldn't navigate to siblings as it currently does. (#116 (comment))

To clarify, it does collapse/expand and navigate to parent/child next- until it reaches a dead end, then it moves to the next branch. I thought this was a nice to have. Sounds like you don't?

Looks like Chrome behaves the way you describe too, so I don't object to changing the behavior as you describe.

gaearon:
> I thought this was a nice to have. Sounds like you don't?

I don't like it because it makes fast repeated presses less reliable. For example I can't just long-press "right" to get to the first deepest child of something. I need to stop carefully or it'll eventually take me outside the element I started with.

gaearon:
>Our key down handler in the Tree ignores events coming from an <input> (since that would be annoying in most cases). The click-to-inspect button is a toggle (checkbox), so it causes subsequent key down events to be ignored until something else is focused.

Should we put focus in the tree right after successful selection? Sounds to me like this is the desired behavior.

gaearon:
Here's a fix for the focus thing. https://github.com/bvaughn/react-devtools-experimental/pull/123

bvaughn:
> I don't like it because it makes fast repeated presses less reliable.

Makes sense. I wondered if you were using it that way.

> Should we put focus in the tree right after successful selection? Sounds to me like this is the desired behavior.

Yup, sounds right.

PR #123, opened 2019-04-10
Picking a DOM node focuses the tree

gaearon:
Addresses part of https://github.com/bvaughn/react-devtools-experimental/issues/121.
This matches Chrome behavior.


bvaughn:
Got a conflict now after the focus style change.

PR #124, opened 2019-04-10
Left and right arrows only traverse parents

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/121#issuecomment-481773140.

### DIFF
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -134,7 +134,11 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
           const selectedElement = store.getElementAtIndex(
             ((selectedElementIndex: any): number)
           );
-          if (selectedElement !== null && selectedElement.children.length > 0) {
+          if (
+            selectedElement !== null &&
+            selectedElement.children.length > 0 &&
+            !selectedElement.isCollapsed
+          ) {
             const firstChildID = selectedElement.children[0];
             const firstChildIndex = store.getIndexOfElementID(firstChildID);
             if (firstChildIndex !== null) {
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -40,6 +40,7 @@ type Context = {|
   selectedElementID: number | null,
   selectedElementIndex: number | null,
   getElementAtIndex(index: number): Element | null,
+  selectChildElementInTree(): void,
   selectElementAtIndex(index: number): void,
   selectElementByID(id: number | null): void,
   selectNextElementInTree(): void,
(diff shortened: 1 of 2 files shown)


gaearon:
Hmm actually this seems insufficient.

gaearon:
Should be good now

src/devtools/views/Components/TreeContext.js
```diff
@@ -127,6 +129,24 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
...
         break;
+      case 'SELECT_CHILD_ELEMENT_IN_TREE':
+        if (selectedElementIndex !== null) {
+          const selectedElement = store.getElementAtIndex(
+            ((selectedElementIndex: any): number)
+          );
+          if (
+            selectedElement !== null &&
+            selectedElement.children.length > 0 &&
+            !selectedElement.isCollapsed
```
bvaughn: I don't think we need this check, and maybe we should remove it to be more consistent with the other actions here. (The layout effect below will ensure that if we were to select a child within a collapsed subtree, the full tree above the child would be expanded.)

```diff
diff --git a/src/devtools/views/Components/TreeContext.js b/src/devtools/views/Components/TreeContext.js
index d610cdd..d40802f 100644
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -136,12 +136,14 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
           );
           if (
             selectedElement !== null &&
-            selectedElement.children.length > 0 &&
-            !selectedElement.isCollapsed
+            selectedElement.children.length > 0
           ) {
             const firstChildID = selectedElement.children[0];
             const firstChildIndex = store.getIndexOfElementID(firstChildID);
             if (firstChildIndex !== null) {
+              lookupIDForIndex = false;
+
+              selectedElementID = firstChildID;
               selectedElementIndex = firstChildIndex;
             }
           }
```

Thoughts?

src/devtools/views/Components/TreeContext.js
```diff
@@ -157,8 +177,12 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
             ((selectedElementIndex: any): number)
           );
           if (selectedElement !== null && selectedElement.parentID !== null) {
-            selectedElementIndex =
-              store.getIndexOfElementID(selectedElement.parentID) || 0;
+            const parentIndex = store.getIndexOfElementID(
+              selectedElement.parentID
+            );
+            if (parentIndex !== null) {
+              selectedElementIndex = parentIndex;
```
bvaughn: nit: I guess we could avoid the unnecessary index-to-element lookup below here too?

```js
if (parentIndex !== null) {
  lookupIDForIndex = false;

  selectedElementID = selectedElement.parentID;
  selectedElementIndex = parentIndex;
}
```

Dunno that it's necessarily an important change.

src/devtools/views/Components/TreeContext.js
gaearon: I only added this check to fix the bug I mentioned to you earlier. But as you see from next commit that bug might have been caused by a missing dependency. I’ll double check.

bvaughn: Ah, ok. I tested the diff above and it seemed okay 😄

gaearon: The bug only showed up on large and complex internal trees.

bvaughn: Intriguing.

PR #126, opened 2019-04-10
Ignore hover when navigating with keyboard

gaearon:
This cleans up the selection/hover behavior when using the keyboard.
Two main changes:

* Hover while we're using the keyboard navigation is hidden. It reappears on mousemove.
* When we're using the keyboard, the DOM selection follows the selected item.

This matches Chrome behavior. (Except that Chrome also keeps last hovered item *always* visible but doesn't update it during keyboard nav — which I think is actually worse than what this PR does.)

Screen Recording 2019-04-10 at 08 22 PM

### DIFF
--- a/src/devtools/views/Components/Tree.js
+++ b/src/devtools/views/Components/Tree.js
@@ -27,7 +27,9 @@ export type ItemData = {|
   baseDepth: number,
   numElements: number,
   getElementAtIndex: (index: number) => Element | null,
+  isNavigatingWithKeyboard: boolean,
   lastScrolledIDRef: { current: number | null },
+  onElementMouseEnter: (id: number) => void,
   treeFocused: boolean,
 |};
 
@@ -49,6 +51,9 @@ export default function Tree(props: Props) {
   } = useContext(TreeContext);
   const bridge = useContext(BridgeContext);
   const store = useContext(StoreContext);
+  const [isNavigatingWithKeyboard, setIsNavigatingWithKeyboard] = useState(
+    false
+  );
   // $FlowFixMe https://github.com/facebook/flow/issues/7341
   const listRef = useRef<FixedSizeList<ItemData> | null>(null);
   const treeRef = useRef<HTMLDivElement | null>(null);
@@ -100,8 +105,6 @@ export default function Tree(props: Props) {
       }
 
       let element;
-
-      // eslint-disable-next-line default-case
       switch (event.key) {
         case 'ArrowDown':
           event.preventDefault();
@@ -143,7 +146,10 @@ export default function Tree(props: Props) {
           event.preventDefault();
           selectPreviousElementInTree();
           break;
+        default:
+          return;
       }
+      setIsNavigatingWithKeyboard(true);
     };
 
     // It's important to listen to the ownerDocument to support the browser extension.
@@ -163,8 +169,8 @@ export default function Tree(props: Props) {
     store,
   ]);
 
+  // Focus management.
   const handleBlur = useCallback(() => setTreeFocused(false));
-
   const handleFocus = useCallback(() => {
     setTreeFocused(true);
 
@@ -189,6 +195,53 @@ export default function Tree(props: Props) {
     [selectedElementID, selectOwner]
   );
 
+  const highlightElementInDOM = useCallback(
+    (id: number) => {
+      const element = store.getElementByID(id);
+      const rendererID = store.getRendererIDForElement(id);
+      if (element !== null) {
+        bridge.send('highlightElementInDOM', {
+          displayName: element.displayName,
+          hideAfterTimeout: false,
+          id,
+          rendererID,
+          scrollIntoView: false,
+        });
+      }
+    },
+    [store, bridge]
+  );
+
+  // If we switch the selected element while using the keyboard,
+  // start highlighting it in the DOM instead of the last hovered node.
+  useEffect(() => {
+    if (isNavigatingWithKeyboard && selectedElementID !== null) {
+      highlightElementInDOM(selectedElementID);
+    }
+  }, [isNavigatingWithKeyboard, highlightElementInDOM, selectedElementID]);
+
+  // Highlight last hovered element.
+  const handleElementMouseEnter = useCallback(
+    id => {
+      // Ignore hover while we're navigating with keyboard.
+      // This avoids flicker from the hovered nodes under the mouse.
+      if (!isNavigatingWithKeyboard) {
+        highlightElementInDOM(id);
+      }
+    },
+    [isNavigatingWithKeyboard, highlightElementInDOM]
+  );
+
+  const handleMouseMove = useCallback(() => {
+    // We started using the mouse again.
+    // This will enable hover styles in individual rows.
+    setIsNavigatingWithKeyboard(false);
+  }, []);
+
+  const handleMouseLeave = useCallback(() => {
+    bridge.send('clearHighlightedElementInDOM');
+  }, [bridge]);
+
   // Let react-window know to re-render any time the underlying tree data changes.
   // This includes the owner context, since it controls a filtered view of the tree.
   const itemData = useMemo<ItemData>(
@@ -196,15 +249,21 @@ export default function Tree(props: Props) {
       baseDepth,
       numElements,
       getElementAtIndex,
+      isNavigatingWithKeyboard,
+      onElementMouseEnter: handleElementMouseEnter,
       lastScrolledIDRef,
       treeFocused,
     }),
-    [baseDepth, numElements, getElementAtIndex, lastScrolledIDRef, treeFocused]
+    [
+      baseDepth,
+      numElements,
+      getElementAtIndex,
+      isNavigatingWithKeyboard,
+      handleElementMouseEnter,
+      lastScrolledIDRef,
+      treeFocused,
+    ]
   );
-
-  const handleMouseLeave = useCallback(() => {
-    bridge.send('clearHighlightedElementInDOM');
-  }, [bridge]);
 
   return (
     <div className={styles.Tree} ref={treeRef}>
@@ -217,6 +276,7 @@ export default function Tree(props: Props) {
         onBlur={handleBlur}
         onFocus={handleFocus}
         onKeyPress={handleKeyPress}
+        onMouseMove={handleMouseMove}
         onMouseLeave={handleMouseLeave}
         ref={focusTargetRef}
         tabIndex={0}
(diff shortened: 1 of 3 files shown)


gaearon:
Pushed a better implementation that keeps hover state local. Should be ready for review.

src/devtools/views/Components/Tree.js
```diff
@@ -189,23 +195,76 @@ export default function Tree(props: Props) {
...
+    }
+  }, [isNavigatingWithKeyboard, highlightElementInDOM, selectedElementID]);
+
+  // Highlight last hovered element.
+  const handleElementMouseEnter = useCallback(
+    id => {
+      // Ignore hover while we're navigating with keyboard.
+      // This avoids flicker from the hovered nodes under the mouse.
+      if (!isNavigatingWithKeyboard) {
+        highlightElementInDOM(id);
```
bvaughn: I understand _why_ this is the case, but it seems unfortunately that we send the "_highlightElementInDOM_" message at the time of the mouse-enter event, but we wait until the commit phase to send it in the case of keyboard navigation. I wish those two could be collocated.

Maybe if we elevated the hover state to the `Tree` (e.g. `hoveredElementID`) then we could use the same effect to handle both cases. Would this be worth doing?

bvaughn: I'm not convinced this is better. Just sharing for discussion purposes.
```diff
diff --git a/src/devtools/views/Components/Element.js b/src/devtools/views/Components/Element.js
index 16fb243..3bcaf64 100644
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -7,7 +7,6 @@ import React, {
   useLayoutEffect,
   useMemo,
   useRef,
-  useState,
 } from 'react';
 import { ElementTypeClass, ElementTypeFunction } from 'src/devtools/types';
 import Store from 'src/devtools/store';
@@ -28,10 +27,10 @@ type Props = {
 };
 
 export default function ElementView({ data, index, style }: Props) {
-  const [isHovered, setIsHovered] = useState(false);
   const {
     baseDepth,
     getElementAtIndex,
+    hoveredElementID,
     ownerStack,
     selectOwner,
     selectedElementID,
@@ -48,6 +47,7 @@ export default function ElementView({ data, index, style }: Props) {
     onElementMouseEnter,
   } = data;
   const id = element === null ? null : element.id;
+  const isHovered = hoveredElementID === id;
   const isSelected = selectedElementID === id;
 
   const handleDoubleClick = useCallback(() => {
@@ -94,16 +94,11 @@ export default function ElementView({ data, index, style }: Props) {
   );
 
   const handleMouseEnter = useCallback(() => {
-    setIsHovered(true);
     if (id !== null) {
       onElementMouseEnter(id);
     }
   }, [onElementMouseEnter, id]);
 
-  const handleMouseLeave = useCallback(() => {
-    setIsHovered(false);
-  }, []);
-
   // Handle elements that are removed from the tree while an async render is in progress.
   if (element == null) {
     console.warn(`<ElementView> Could not find element at index ${index}`);
@@ -130,7 +125,6 @@ export default function ElementView({ data, index, style }: Props) {
     <div
       className={className}
       onMouseEnter={handleMouseEnter}
-      onMouseLeave={handleMouseLeave}
       onMouseDown={handleMouseDown}
       onDoubleClick={handleDoubleClick}
       style={{
diff --git a/src/devtools/views/Components/Tree.js b/src/devtools/views/Components/Tree.js
index dd3fe62..85596a3 100644
--- a/src/devtools/views/Components/Tree.js
+++ b/src/devtools/views/Components/Tree.js
@@ -39,6 +39,7 @@ export default function Tree(props: Props) {
   const {
     baseDepth,
     getElementAtIndex,
+    hoveredElementID,
     numElements,
     ownerStack,
     selectedElementID,
@@ -49,6 +50,7 @@ export default function Tree(props: Props) {
     selectOwner,
     selectParentElementInTree,
     selectPreviousElementInTree,
+    setHoveredElementByID,
   } = useContext(TreeContext);
   const bridge = useContext(BridgeContext);
   const store = useContext(StoreContext);
@@ -210,13 +212,21 @@ export default function Tree(props: Props) {
     [store, bridge]
   );
 
-  // If we switch the selected element while using the keyboard,
-  // start highlighting it in the DOM instead of the last hovered node.
+  // Update the highlighted DOM element based on the current selection state.
   useEffect(() => {
-    if (isNavigatingWithKeyboard && selectedElementID !== null) {
-      highlightElementInDOM(selectedElementID);
+    if (isNavigatingWithKeyboard) {
+      if (selectedElementID !== null) {
+        highlightElementInDOM(selectedElementID);
+      }
+    } else if (hoveredElementID !== null) {
+      highlightElementInDOM(hoveredElementID);
     }
-  }, [isNavigatingWithKeyboard, highlightElementInDOM, selectedElementID]);
+  }, [
+    hoveredElementID,
+    isNavigatingWithKeyboard,
+    highlightElementInDOM,
+    selectedElementID,
+  ]);
 
   // Highlight last hovered element.
   const handleElementMouseEnter = useCallback(
@@ -224,10 +234,10 @@ export default function Tree(props: Props) {
       // Ignore hover while we're navigating with keyboard.
       // This avoids flicker from the hovered nodes under the mouse.
       if (!isNavigatingWithKeyboard) {
-        highlightElementInDOM(id);
+        setHoveredElementByID(id);
       }
     },
-    [isNavigatingWithKeyboard, highlightElementInDOM]
+    [isNavigatingWithKeyboard, setHoveredElementByID]
   );
 
   const handleMouseMove = useCallback(() => {
diff --git a/src/devtools/views/Components/TreeContext.js b/src/devtools/views/Components/TreeContext.js
index d610cdd..5c7cf80 100644
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -36,6 +36,7 @@ import type { Element } from './types';
 type Context = {|
   // Tree
   baseDepth: number,
+  hoveredElementID: number | null,
   numElements: number,
   selectedElementID: number | null,
   selectedElementIndex: number | null,
@@ -46,6 +47,7 @@ type Context = {|
   selectNextElementInTree(): void,
   selectParentElementInTree(): void,
   selectPreviousElementInTree(): void,
+  setHoveredElementByID(id: number | null): void,
 
   // Search
   searchIndex: number | null,
@@ -70,6 +72,7 @@ TreeContext.displayName = 'TreeContext';
 
 type State = {|
   // Tree
+  hoveredElementID: number | null,
   baseDepth: number,
   numElements: number,
   selectedElementID: number | null,
@@ -99,6 +102,7 @@ type Action = {|
     | 'SELECT_PARENT_ELEMENT_IN_TREE'
     | 'SELECT_PREVIOUS_ELEMENT_IN_TREE'
     | 'SELECT_OWNER'
+    | 'SET_HOVERED_ELEMENT_ID'
     | 'SET_SEARCH_TEXT',
   payload?: any,
 |};
@@ -107,6 +111,7 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
   const { type, payload } = action;
 
   let {
+    hoveredElementID,
     numElements,
     ownerStack,
     selectedElementIndex,
@@ -193,6 +198,9 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
           selectedElementIndex--;
         }
         break;
+      case 'SET_HOVERED_ELEMENT_ID':
+        hoveredElementID = payload;
+        break;
       default:
         // React can bailout of no-op updates.
         return state;
@@ -213,6 +221,7 @@ function reduceTreeState(store: Store, state: State, action: Action): State {
   return {
     ...state,
 
+    hoveredElementID,
     numElements,
     selectedElementIndex,
     selectedElementID,
@@ -597,6 +606,7 @@ function TreeContextController({ children, viewElementSource }: Props) {
         case 'SELECT_PARENT_ELEMENT_IN_TREE':
         case 'SELECT_PREVIOUS_ELEMENT_IN_TREE':
         case 'SELECT_OWNER':
+        case 'SET_HOVERED_ELEMENT_ID':
         case 'SET_SEARCH_TEXT':
           state = reduceTreeState(store, state, action);
           state = reduceSearchState(store, state, action);
@@ -612,6 +622,7 @@ function TreeContextController({ children, viewElementSource }: Props) {
   const [state, dispatch] = useReducer(reducer, {
     // Tree
     baseDepth: 0,
+    hoveredElementID: null,
     numElements: store.numElements,
     selectedElementIndex: null,
     selectedElementID: null,
@@ -681,11 +692,17 @@ function TreeContextController({ children, viewElementSource }: Props) {
     (id: number) => dispatch({ type: 'SELECT_OWNER', payload: id }),
     [dispatch]
   );
+  const setHoveredElementByID = useCallback(
+    (id: number | null) =>
+      dispatch({ type: 'SET_HOVERED_ELEMENT_ID', payload: id }),
+    [dispatch]
+  );
 
   const value = useMemo(
     () => ({
       // Tree (derived from Store or owners state)
       baseDepth: state.baseDepth,
+      hoveredElementID: state.hoveredElementID,
       numElements: state.numElements,
       selectedElementID: state.selectedElementID,
       selectedElementIndex: state.selectedElementIndex,
@@ -696,6 +713,7 @@ function TreeContextController({ children, viewElementSource }: Props) {
       selectNextElementInTree,
       selectParentElementInTree,
       selectPreviousElementInTree,
+      setHoveredElementByID,
 
       // Search
       searchIndex: state.searchIndex,
@@ -726,6 +744,7 @@ function TreeContextController({ children, viewElementSource }: Props) {
       selectParentElementInTree,
       selectOwner,
       selectPreviousElementInTree,
+      setHoveredElementByID,
       setSearchText,
       state,
       viewElementSource,
```

gaearon: That’s what I started this PR with :P but then decided to change because it seemed excessive to re-render all items when just one of them is hovered.

bvaughn: 😆I don't read individual commits, so I didn't know.

But hey, that's fair.

bvaughn: Okay. Let's forget about that possible idea then. I agree that it would be better to avoid re-rendering items because of hover.

The added code complexity kind of bums me here still. Are we sure it's worth it?

Chrome Elements tab doesn't do this, for what it's worth.

gaearon: Doesn’t do what exactly? It does follow the keyboard.

The only thing I see it not doing is that it keeps last registered hover visible. But it definitely ignores hover while using the keyboard.

bvaughn: > Doesn’t do what exactly? It does follow the keyboard.

It doesn't suppress the hover highlight.

I see what you mean about master not keeping the hover highlight fixed/sticky though.

Okay. I guess it's fine then. 😄

PR #133, opened 2019-04-10
Automated tests [WIP]

bvaughn:
Relates to #115

Test goals:
* Write test code using a real React renderer (e.g. ReactDOM) injected into the real DevTools hook.
* Use the real renderer interface to generate tree op codes and send them through a real `Bridge` to a real `Store`.
* Use Jest snapshot mechanism with a pretty-printed version of the `Store` tree.

So a test like this:
```js
const Grandparent = () => (
  <React.Fragment>
    <Parent />
    <Parent />
  </React.Fragment>
);
const Parent = () => [<Child key="one" />, <Child key="two" />];
const Child = () => <div>Hi!</div>;

ReactDOM.render(<Grandparent />, document.createElement("div"));

expect(store).toMatchSnapshot();
```

Will generate a snapshot like this:
```
[root]
  ▾ <Grandparent>
    ▾ <Parent>
        <Child key="one">
        <Child key="two">
    ▾ <Parent>
        <Child key="one">
        <Child key="two">
```


gaearon:
😍😍😍

bvaughn:
Okay Dan. Maybe take a look at this and let me know your initial thoughts.

There's a lot more to do:
* Lots of additional tests cases to write for the `Store`
* Tests to add for the `DevTools` UI itself (to cover contexts and effects)

But I think it's at a point where I'd be interested in your initial feedback. 😄

bvaughn:
I'm going to go ahead and land this as a starting point so we have _some_ tests and CI is running, etc.

I'll be happy to iterate and make changes based on feedback if you have any suggestions though!

PR #134, opened 2019-04-10
Improve bridge batching logic

### LINKED ISSUE #132: Tree view tears across updates
Minor issue, but IMO still a bug…

Repro:

1. Go to https://react-devtools-experimental.now.sh/ and click "unmount test app" and "mount test app" repeatedly

Expected: With each click, the devtools tree view updates exactly once.
Actual: In most cases, you can see the List component appear or disappear before the others do, causing two visual updates in quick succession.

Since they are rendered in the same tick (and there is exactly one paint to the screen for the components themselves), I expect the devtools to paint once too.

(Tried to capture a GIF but it's hard to grab since it's a fast flicker.)

sophiebits:
Fixes #132.

### DIFF
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -13,7 +13,6 @@ type Message = {|
 
 export default class Bridge extends EventEmitter {
   _messageQueue: Array<any> = [];
-  _time: number | null = null;
   _timeoutID: TimeoutID | null = null;
   _wall: Wall;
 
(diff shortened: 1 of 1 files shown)


sophiebits:
Also fixes an O(n^2) issue with the splice-in-a-loop.

bvaughn:
Thanks Sophie! 😄

Issue #139, opened 2019-04-12
Navigating search results should highlight the next match on the page


gaearon:
Similar to https://github.com/bvaughn/react-devtools-experimental/pull/126, we should highlight the matching nodes as you press "Enter" on search results. Otherwise it's hard to tell what search matches correspond to.

bvaughn:
This was fixed by your PR #140. GitHub just got confused because of how I rebased and merged it. Closing!

PR #140, opened 2019-04-12
Track search navigation in DOM

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/139.

This makes it significantly easier to understand what the search matches represent on the screen. E.g. you can search for a Relay container internally, press Enter many times and quickly see different parts highlight.

It reuses the mechanism I introduced in https://github.com/bvaughn/react-devtools-experimental/pull/126. If you navigate the tree via something else than scrolling (keyboard or search) then we want to hide hover and to start highlighting the DOM components in response to those changes.

If you're super concerned about this component growing, I can try to pull it out into a custom Hook.

Screen Recording 2019-04-12 at 06 58 PM

Screen Recording 2019-04-12 at 07 05 PM

### DIFF
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -44,7 +44,7 @@ export default function ElementView({ data, index, style }: Props) {
   const {
     lastScrolledIDRef,
     treeFocused,
-    isNavigatingWithKeyboard,
+    isUsingKeyboardOrSearch,
     onElementMouseEnter,
   } = data;
   const id = element === null ? null : element.id;
@@ -122,7 +122,7 @@ export default function ElementView({ data, index, style }: Props) {
     className = treeFocused
       ? styles.SelectedElement
       : styles.InactiveSelectedElement;
-  } else if (isHovered && !isNavigatingWithKeyboard) {
+  } else if (isHovered && !isUsingKeyboardOrSearch) {
     className = styles.HoveredElement;
   }
 
--- a/src/devtools/views/Components/SearchInput.js
+++ b/src/devtools/views/Components/SearchInput.js
@@ -8,9 +8,12 @@ import Icon from '../Icon';
 
 import styles from './SearchInput.css';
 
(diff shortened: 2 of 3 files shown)


gaearon:
I haven’t checked but I was worried about cases where some change to the store (not necessarily by the user) would make it appear as if the search was used — but it was just due to some other update that happened to affect the search state.

I also think it doesn’t necessarily need to change the results in order to select. For example where there’s 1/1 result, I’d still expect clicking “enter” after picking a different node to select the searched one. This is what tells me it’s more about the last *interaction* than the store state.

gaearon:
I can definitely change it though. But in my experience performing side effects far from what causes them (especially when there’s a store in the middle) can lead to unintended consequence. It’s harder to tell the intent of a state change. I think our selection logic suffers from that too — it’s harder to know when and how far to autoscroll because it’s hard to tell a user-initiated select by click from a more “automatic” selection like by search or from a DOM node. By the time it becomes a state change (or no change), the information about intent is lost.

gaearon:
I'm happy to defer to you on this. Please feel free to add your change on top if you feel this is better. We can see if we find any reasons to do otherwise later.

bvaughn:
> But in my experience performing side effects far from what causes them (especially when there’s a store in the middle) can lead to unintended consequence.

I think this one deserves a little more thought. Let's let it sit for the weekend?

It bothers me a little that we're passing around callbacks to observe changes in things that context is already notifying us of.

gaearon:
Your suggestion does seem simpler for now so I'd just roll with it and see if we learn anything from this. Nothing jumps out at me.

bvaughn:
Ok 👍

gaearon:
How to apply a diff? 😄

bvaughn:
lol I'll just  stash pop it locally

bvaughn:
Huh. I rebased  before merging and GitHub seems to be confused, but it's been merged. 👍

PR #145, opened 2019-04-12
Auto-update Chrome extension

bvaughn:
Make use of the [`update_url`](https://developers.chrome.com/apps/autoupdate) mechanism to push  updates to beta testers.


bvaughn:
This is a little janky, but it's probably okay for now.

Issue #142, opened 2019-04-12
Horizontal autoscroll on search should include the whole tag name


gaearon:
As far as I can see (at least in my testing on FB), it just ensures the beginning of the tag is seen horizontally when we search and the item is deeply nested. This means a search often brings up an almost completely blank screen, with the matching result barely visible at the very right edge. This seems bad. Instead, ideally, I think the search should center the search result horizontally if it wasn't in the observable screen — or at least ensure the whole tag name fits.

gaearon:
I can also reproduce this on the work laptop. Not personal one. Not sure why. I'll dig in later.

gaearon:
Oohh interesting. It's because the current strategy doesn't do anything if even a *bit* of the name is visible. But whether a bit is visible or not depends on the previous horizontal scroll bar offset. So it's pretty random depending on which part of the tree you were inspecting before.

PR #148, opened 2019-04-13
Consistently scroll component name into view

gaearon:
This has been bugging me for the last two weeks. I thought there's no good fix but this one works.

## Before

Screen Recording 2019-04-13 at 01 59 am

## After

Screen Recording 2019-04-13 at 01 59 am (1)

## What Else?

This also fixes https://github.com/bvaughn/react-devtools-experimental/issues/142. You can try it by inspecting FB trees with "find by DOM node". Previously, it would occasionally fail to fully display the whole component name. Now it always does.

### DIFF
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -80,15 +80,17 @@ export default function ElementView({ data, index, style }: Props) {
       // However, even calling scrollIntoView() on their parent node
       // wouldn't guarantee that it will be *fully* brought into view.
       // As a workaround, we'll have two anchor spans, and scroll each into view.
-      if (scrollAnchorStartRef.current !== null) {
-        scrollAnchorStartRef.current.scrollIntoView({
+      if (scrollAnchorEndRef.current !== null) {
+        scrollAnchorEndRef.current.scrollIntoView({
           behavior: 'auto',
           block: 'nearest',
           inline: 'nearest',
         });
       }
-      if (scrollAnchorEndRef.current !== null) {
-        scrollAnchorEndRef.current.scrollIntoView({
+      if (scrollAnchorStartRef.current !== null) {
+        // We scroll the start anchor last because it's
+        // more important for it to be in the view.
+        scrollAnchorStartRef.current.scrollIntoView({
           behavior: 'auto',
           block: 'nearest',
           inline: 'nearest',
--- a/src/devtools/views/Components/Element.js
+++ b/src/devtools/views/Components/Element.js
@@ -56,7 +56,8 @@ export default function ElementView({ data, index, style }: Props) {
     }
   }, [id, selectOwner]);
 
-  const ref = useRef<HTMLSpanElement | null>(null);
+  const scrollAnchorStartRef = useRef<HTMLSpanElement | null>(null);
+  const scrollAnchorEndRef = useRef<HTMLSpanElement | null>(null);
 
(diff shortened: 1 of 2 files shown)


src/devtools/views/Components/Element.js
```diff
@@ -149,10 +162,11 @@ export default function ElementView({ data, index, style }: Props) {
         marginBottom: `-${style.height}px`,
       }}
     >
+      <span className={styles.ScrollAnchor} ref={scrollAnchorStartRef} />
```
bvaughn: If we're going to at least one extra `<span>` tag for each row,  why not just wrap one  around the toggle+name+$r group? Seems like maybe that would be a  simpler fix? (Unless I'm missing something?)

gaearon: The comment mentions this doesn't work. :-)

It's not sufficient because `scrollIntoView` doesn't actually guarantee its contents gets scrolled into view. (Yeah, wtf.) For some reason it's happy when some parts of the content are not in view — as long as some _are_ in view. I don't know what exact browser heuristics are. But that's the reason #142 is already happening.

By using 0px width we trick the browser to not apply these heuristics, and make the "is in view" decision "binary". Either it is, or it isn't. We do this for both sides, getting the desired behavior.

In fact you can see `scrollIntoView` alone isn't sufficient just from the "before" gif. Because we already have it around the component name — and yet you can see the component name being cut off a little all the time.

bvaughn: Oh, I misunderstood the comment to mean that scrolling the whole row didn't work, because it's larger than the content. I thought a tightly-fit `<span>` _would_ work.  My apologies.

> It's not sufficient because scrollIntoView doesn't actually guarantee its contents gets scrolled into view.

Wtf indeed

gaearon: Browsers ¯\_(ツ)_/¯

bvaughn: To clarify, because I'm still wtf-ing hard over here. The comment says:

> // However, even calling scrollIntoView() on their parent node
> // wouldn't guarantee that it will be *fully* brought into view.

Their "parent node" is literally the whole row:
Screen Shot 2019-04-12 at 6 22 52 PM

I was thinking that a `<span>` wrapped around them (within the row, but not `width:100%`) would also fix this.

I assume you tried that and it didn't, but I just wanted to rule out a possible miscommunication 😄

bvaughn: > In fact you can see scrollIntoView alone isn't sufficient just from the "before" gif. Because we already have it around the component name

I'm pretty sure this answers my question above

gaearon: Maybe I tried it wrong. (Yes, I tried it.)

This doesn’t rule out my original point though. Which is — if scrollIntoView guaranteed that a node fully gets into view, we wouldn’t observe the existing behavior where the component name *already doesn’t*.

So I’m not sure how even in theory wrapping them would help.

bvaughn: Yeah, that bit popped back into my head after I'd left the comment. (I was on my way to the barber.) So I guess `scrollIntoView` is just...busted. Makes no sense though.

gaearon:
I pushed an additional fix that makes sure that when *only one* of the anchors can be visible (e.g. a really long component name), we prefer the start one.

PR #149, opened 2019-04-13
Don't reset selected item on search mismatch or exit

gaearon:
When I exit the search mode, this keeps the last selected item. There's no use resetting it — and usually I want to continue from that point on.

Similarly, if the search no longer matches, this keeps the last selected item. I don't think resetting it does any good. On the opposite, it often leads to me pressing "down" and accidentally ending up on the very top of the tree.

Screen Recording 2019-04-13 at 04 18 pm

### DIFF
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -369,16 +369,11 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
       searchIndex = newSearchIndex;
     }
   }
-  if (didRequestSearch) {
-    if (searchIndex === null) {
-      selectedElementIndex = null;
-      selectedElementID = null;
-    } else {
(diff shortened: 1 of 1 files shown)


src/devtools/views/Components/TreeContext.js
```diff
@@ -369,16 +369,11 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
       searchIndex = newSearchIndex;
     }
   }
-  if (didRequestSearch) {
-    if (searchIndex === null) {
-      selectedElementIndex = null;
-      selectedElementID = null;
```
gaearon: Effectively I just removed this branch.

bvaughn: Yeah, [`?w=1`](https://github.com/bvaughn/react-devtools-experimental/pull/149/files?w=1) makes that clear 😁

Issue #143, opened 2019-04-12
Misalignment of borders


gaearon:
Look at the borders closely:

* There's an unwanted horizontal gap between the left and right pane border close to the "Inspect" button
* The horizontal lines don't line up

<img width="558" alt="Screen Shot 2019-04-12 at 21 01 13" src="https://user-images.githubusercontent.com/810438/56063059-262c5e00-5d66-11e9-959f-181e6831fee6.png">

bvaughn:
Fixed in b2cf896. I had just fat-fingered where the padding was when I re-added it.

PR #152, opened 2019-04-13
Remove broken padding from Tree

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/143.

It kinda made sense when we had a focus ring but I don't think it's desirable to have anymore either.

## Before

Screen Shot 2019-04-13 at 16 55 48

## After

Screen Shot 2019-04-13 at 16 57 56

### DIFF
--- a/src/devtools/views/Components/Tree.css
+++ b/src/devtools/views/Components/Tree.css
@@ -4,7 +4,6 @@
   display: flex;
   flex-direction: column;
   border-top: 1px solid var(--color-border);
-  padding: 0.25rem;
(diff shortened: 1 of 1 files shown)


bvaughn:
I actually think the padding is nice to have. The tree feels cramped otherwise (the selected item border looks weird with a corner radius but also bumping up against the edge)

I'd rather fix the border pixel alignment issue without removing the padding. I'll just do that.

bvaughn:
Fixed in b2cf896

Issue #153, opened 2019-04-13
Selection width isn't adjusted on window resize


gaearon:
![Screen Recording 2019-04-13 at 05 02 pm](https://user-images.githubusercontent.com/810438/56082219-10727380-5e0e-11e9-9b8b-0e6d136fa8d8.gif)

PR #154, opened 2019-04-13
Adjust highlighting on window resize

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/153.

Instead of a ref that's being read in a render phase, I re-modeled this remembered "minimum width" as state. This state is derived. It gets reset when we enter the owner view — or if it's resized. If this state is reset, we do a render with `100%` and then fill in the state again in the effect.

Screen Recording 2019-04-13 at 06 26 pm

### DIFF
--- a/src/devtools/views/Components/Tree.js
+++ b/src/devtools/views/Components/Tree.js
@@ -333,22 +333,32 @@ function InnerElementType({ style, ...rest }) {
   // What we can do instead, is passively measure the width of the current rows,
   // and ensure that once we've grown to a new max size, we don't shrink below it.
   // This improves the user experience when scrolling between wide and narrow rows.
-  // We shouldn't retain this width across different conceptual trees though,
-  // so when the user opens the "owners tree" view, we should discard the previous width.
   const divRef = useRef<HTMLDivElement | null>(null);
-  const minWidthRef = useRef<number | null>(null);
-  const minWidth =
-    ownerStack.length > 0 || minWidthRef.current === null
-      ? '100%'
-      : minWidthRef.current;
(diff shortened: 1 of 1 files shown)


gaearon:
I guess one downside of this particular solution is that resizing would reset the "stretched" width. That's probably fine. But maybe we can have a way to preserve it too?

src/devtools/views/Components/Tree.js
```diff
@@ -333,23 +333,33 @@ function InnerElementType({ style, ...rest }) {
...
-  // We shouldn't retain this width across different conceptual trees though,
-  // so when the user opens the "owners tree" view, we should discard the previous width.
   const divRef = useRef<HTMLDivElement | null>(null);
-  const minWidthRef = useRef<number | null>(null);
-  const minWidth =
-    ownerStack.length > 0 || minWidthRef.current === null
-      ? '100%'
-      : minWidthRef.current;
+  const [minWidth, setMinWidth] = useState(null);
   useEffect(() => {
```
bvaughn: This one's failing lint now.

```
src/devtools/views/Components/Tree.js
  339:3  error  React Hook useEffect contains a call to 'setMinWidth'. Without a list of dependencies, this can lead to an infinite chain of updates. To fix this, pass [] as a second argument to the useEffect Hook  react-hooks/exhaustive-deps
```

gaearon: Let's disable this line for now. The warning is fair but I don't know of a better solution yet.

bvaughn: K ~ https://github.com/bvaughn/react-devtools-experimental/issues/156

Issue #157, opened 2019-04-14
Investigate internal perf issue when toggling Suspense


gaearon:
Profiles show we're spending a lot of time in the `renderer.js` code. Curious if there's some bad loop condition there.

bvaughn:
I'd be curious how it compares to the legacy DevTools. (Wondering if it's necessarily slow because of how we have to walk-and-diff or if we're doing something dumb)

gaearon:
First look: we're spending a *lot* of time messing with the typed array on mounting a large tree as a result of an update. I wonder if we could avoid re-creating it on every operation, and instead buffer them in a mutable array.

<img width="558" alt="Screen Shot 2019-04-15 at 2 19 29 PM" src="https://user-images.githubusercontent.com/810438/56135887-8ea75500-5f89-11e9-9f60-e3b07205f4c9.png">

gaearon:
First improvement: https://github.com/bvaughn/react-devtools-experimental/pull/159

PR #161, opened 2019-04-15
Don't materialize typed array until the transfer

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/157.

We used to copy and recreate the typed array on every operation. This is very costly.

This changes our queue to be a mutable array of integers — and delays the movement into typed array until the transfer. (Note that in practice we don't get to use it in Chrome, and have to fall back to JSON anyway. But I didn't see significant overhead to creating a final typed array so it seems like we can keep that constraint and enjoy even faster transfer on FF for now.)

### Before

Screen Shot 2019-04-15 at 4 26 19 PM

### After

Screen Shot 2019-04-15 at 4 25 10 PM

### Before

Screen Shot 2019-04-15 at 4 34 03 PM

### After

Screen Shot 2019-04-15 at 4 32 15 PM

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -577,22 +577,34 @@ export function attach(
     );
   }
 
-  let pendingOperations: Uint32Array = new Uint32Array(0);
-  let pendingOperationsQueue: Array<Uint32Array> | null = [];
+  let pendingOperations: Array<number> = [];
+  let pendingOperationsQueue: Array<Array<number>> | null = [];
 
-  function addOperation(
-    newAction: Uint32Array,
-    addToStartOfQueue: boolean = false
-  ): void {
-    const oldActions = pendingOperations;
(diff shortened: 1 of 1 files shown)


src/backend/renderer.js
```diff
@@ -577,22 +577,34 @@ export function attach(
...
+      }
+    }
+
     if (addToStartOfQueue) {
-      pendingOperations.set(newAction);
-      pendingOperations.set(oldActions, newAction.length);
+      pendingOperations.splice.apply(
+        pendingOperations,
+        [0, 0].concat(nextOperation)
+      );
```
sophiebits: NB: this is still an O(n) operation which makes it O(n^2) to process n "unmount" commands, for example. A double-ended queue is probably the right data structure here if we do need to insert at both the start and end.

PR #168, opened 2019-04-16
Configurable tree default collapsed/expanded behavior

### LINKED ISSUE #167: Bugs when everything is collapsed by default
I wanted to experiment with a UI that's collapsed by default.

But just flipping this to `true`:

https://github.com/bvaughn/react-devtools-experimental/blob/85179fe4a7e7cb2e9d79733eec4839443393e7ba/src/devtools/store.js#L604

creates weird visual artifacts:

Screen Shot 2019-04-16 at 7 48 28 PM

Would be cool to fix it, at least just for experimentation.

bvaughn:
collapse-by-default-Kapture 2019-04-16 at 13 57 52

Resolves #167

### DIFF
--- a/src/__tests__/store-test.js
+++ b/src/__tests__/store-test.js
@@ -187,6 +187,27 @@ describe('Store', () => {
       act(() => store.toggleIsCollapsed(grandparentID, false));
       expect(store).toMatchSnapshot('6: expand Grandparent');
     });
+
+    it('should support reordering of children', () => {
+      const Component = ({ children = null }) => children;
+
+      const Foo = () => [<Component key="0" />];
+      const Bar = () => [<Component key="0" />, <Component key="1" />];
+      const foo = <Foo key="foo" />;
+      const bar = <Bar key="bar" />;
+
+      const container = document.createElement('div');
+
+      act(() =>
+        ReactDOM.render(<Component>{[foo, bar]}</Component>, container)
+      );
+      expect(store).toMatchSnapshot('1: mount');
+
+      act(() =>
+        ReactDOM.render(<Component>{[bar, foo]}</Component>, container)
+      );
+      expect(store).toMatchSnapshot('3: reorder children');
+    });
   });
 
   describe('collapseNodesByDefault:true', () => {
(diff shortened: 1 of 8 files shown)


src/__tests__/store-test.js
```diff
@@ -21,142 +21,324 @@ describe('Store', () => {
...
-    act(() => ReactDOM.render(<Grandparent count={4} />, container));
-    expect(store).toMatchSnapshot('1: mount');
-
-    act(() => ReactDOM.render(<Grandparent count={2} />, container));
-    expect(store).toMatchSnapshot('2: update');
-
-    act(() => ReactDOM.unmountComponentAtNode(container));
-    expect(store).toMatchSnapshot('3: unmount');
+  it('should not allow a root node to be collapsed', () => {
+    // TODO
```
bvaughn: Will commit a follow up shortly with this test.

bvaughn:
Hold off merging this please. Need to investigate a likely bug related to the collapsed-by-default state.

bvaughn:
I think I've identified and fixed some of the collapsed bug cases in the `Store`. Should be ready for a review now.

src/devtools/store.js
```diff
@@ -862,15 +942,29 @@ export default class Store extends EventEmitter {
...
+            'null'}${key}>${suffix}`
         );
       }
 
       rootWeight += weight;
     });
 
+    // Make sure the pretty-printed test align with the Store's reported number of total rows.
+    if (rootWeight !== this._weightAcrossRoots) {
+      throw Error(
```
gaearon: Nice

bvaughn: Yeah, this was my bad. This was masking some bugs our earlier tests would have shown.

src/__tests__/storeStress-test.js
```diff
@@ -18,6 +18,7 @@ describe('StoreStress', () => {
   beforeEach(() => {
     bridge = global.bridge;
     store = global.store;
+    store.collapseNodesByDefault = false;
```
bvaughn: This default preserves the behavior of the `Store` at the time when the snapshot tests were originally written. We may want to add additional stress tests for the collapse-by-default behavior, but I wanted to reduce churn on this PR.

src/devtools/store.js
```diff
@@ -425,22 +448,67 @@ export default class Store extends EventEmitter {
...
+        element.isCollapsed = true;
+
+        const weightDelta = 1 - element.weight;
+
+        let parentElement = ((this._idToElement.get(
+          element.parentID
+        ): any): Element);
+        while (parentElement != null) {
+          parentElement.weight += weightDelta;
+          parentElement = this._idToElement.get(parentElement.parentID);
```
bvaughn: We don't need to bail out on collapsed parents (like we do below) since we only toggle _one node_ in this case.

src/devtools/store.js
```diff
@@ -425,22 +448,67 @@ export default class Store extends EventEmitter {
...
+            ? 1
+            : currentElement.weight;
+          const weightDelta = newWeight - oldWeight;
+
+          let parentElement = ((this._idToElement.get(
+            currentElement.parentID
+          ): any): Element);
+          while (parentElement != null) {
+            parentElement.weight += weightDelta;
+            if (parentElement.isCollapsed) {
```
bvaughn: It's important to bail out on collapsed parents here, because we are going to walk up the tree and expand them too (and we don't want to add their weight multiple times).

bvaughn:
K. We can fix forward if you have suggestions then.

Issue #169, opened 2019-04-17
Unmounting crash in Concurrent Mode


gaearon:
Test case:

```js
function mountTestApp() {
    const container = document.createElement('div');

  ((document.body: any): HTMLBodyElement).appendChild(container);

  containers.push(container);

    let root = ReactDOM.unstable_createRoot(container);

    const A = () => 'a';
    const B = () => 'b';
    const C = () => 'c';
    const X = () => 'x';
    const Y = () => 'y';
    const Z = () => 'z';
    const a = <A key="a" />;
    const b = <B key="b" />;
    const c = <C key="c" />;
    const z = <Z key="z" />;


    const Never = () => {
      throw new Promise(() => {});
    };

    const Root = ({ children }) => {
      return children;
    };

    let steps = []
    window.addEventListener('click', () => {
      steps.shift()()
    })

    // 6. Verify we can update from each step to each step when moving fallback -> primary.
    // Always start with a fresh container and steps[i].
    steps.push(() => {
      root.render(
        <Root>
          <React.Suspense>{a}</React.Suspense>
        </Root>,
      );
    })
      
    steps.push(() => {
      root.render(
        <Root>
          <React.Suspense fallback={a}><Never /></React.Suspense>
        </Root>,
      )
    });

    steps.push(() => {
      root.unmount()
    });
}
```

PR #172, opened 2019-04-17
Fix unmounting of Suspense in Concurrent Mode

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/169.

The gist of the issue is that in Concurrent Mode, the primary Fiber doesn't actually unmount on suspending. It just gets hidden. So it would still reappear as part of an `ADD` or `REMOVE` event in the future.

In Sync Mode we didn't notice this because we'd always get `REMOVE` event for it first.

See individual commits:

* I moved current stress test into `Sync` file
* Added a similar test for CM
* Added a proposed fix

My fix doesn't seem sufficient to fix a crash in the Profiler though. You can reproduce it if you replace `root.unmount()` with `root.render(<div />)` in https://github.com/bvaughn/react-devtools-experimental/issues/169 repro example, and Profile between the first and the last click. Then pressing "navigate to next commit" button on Profiler will crash. However, it crashes in master too. I'd need help with this because I don't fully know how that CommitTreeBuilder should work.


src/devtools/store.js
```diff
@@ -75,6 +75,11 @@ export default class Store extends EventEmitter {
   // When profiling is in progress, operations are stored so that we can later reconstruct past commit trees.
   _isProfiling: boolean = false;
 
+  // These are IDs which were conceptually "removed" from the tree,
+  // but whose Fibers may resurface later either as "add" or "remove" events.
+  // This is what happens to a Suspended "primary" tree in Concurrent Mode.
+  _hiddenIDs: Set<number> = new Set();
```
bvaughn: It seems unfortunate that Suspense-specific logic has leaked into the frontend. Is there a way we could track and model this in the renderer so that the `Store` and `CommitTreeBuilder` don't need to be aware of this particular quirk of Suspense?

bvaughn: To clarify: Maybe we could track the same set of ids in the backend, and send (or not send) mount/unmount events for the hidden fibers (not sure which would be most appropriate since I haven't really digested the essence of this fix yet).

PR #174, opened 2019-04-18
Select tab when the mouse is pressed

robertknight:
Selecting the tab when the mouse is pressed rather than only when it is
released makes tab switching feel faster and also matches how the Chrome dev
tools behave.

The previous tab selection handler which listens for the radio button's
selection has been kept for the benefit of interaction methods which
don't trigger the mousedown handler.


robertknight:
On macOS at least it is a little bit variable how different kinds of tabs behave in different apps, but browser/document tabs and devtools tabs seem to be consistent in this respect between Chrome/Safari/Firefox.

PR #178, opened 2019-04-18
Assert removal happens in the right order

gaearon:
I just spent about an hour chasing bug that was caused by wrong deletion order (the bug was in some new code, not in master).

I was misled by the comment which claims deletions are reported by React in the child-first order — whereas the opposite is true. I've fixed the comment and added an assertion that ensures this is actually the case.

### DIFF
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -760,8 +760,7 @@ export default class Store extends EventEmitter {
 
           this._idToElement.delete(id);
 
-          parentElement = ((this._idToElement.get(parentID): any): Element);
-          if (parentElement == null) {
+          if (parentID === 0) {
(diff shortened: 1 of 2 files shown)


src/devtools/store.js
```diff
@@ -775,7 +774,15 @@ export default class Store extends EventEmitter {
             if (__DEBUG__) {
               debug('Remove', `fiber ${id} from parent ${parentID}`);
             }
-
+            parentElement = ((this._idToElement.get(parentID): any): Element);
+            if (parentElement === undefined) {
+              throw new Error(
```
bvaughn: Nice

gaearon:
I guess I misled myself then! :D

PR #180, opened 2019-04-18
Fix unmounting in Concurrent Mode

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/169.
Alternative to https://github.com/bvaughn/react-devtools-experimental/pull/172.

This adds tests for Concurrent Mode. The tests uncovered that the "recursively remove children" approach was flawed. In particular, a Fiber that gets suspended doesn't necessarily get unmounted first. That's the whole point of Suspense. But for this reason, we might receive `REMOVE` events for Fibers that we already "forgot" about, causing crashes.

The fix assumes that it's actually safe to crawl the "previous current" tree in the commit phase to discover which primary children to consider unmounted. I chatted with @acdlite about this and he seems to agree this can work. In my testing, it does. I'll also look into adding more edge cases to the test suite in the follow-ups.


src/backend/renderer.js
```diff
@@ -828,6 +818,18 @@ export function attach(
     }
   }
 
+  function unmountFiberChildrenRecursively(fiber: Fiber) {
+    if (__DEBUG__) {
+      debug('unmountFiberChildrenRecursively()', fiber);
+    }
+    let child = fiber.child;
+    while (child !== null) {
+      recordUnmount(child);
```
bvaughn: Shouldn't we be checking if it's a fiber that we actually told the store about to begin with before recording an unmount?

We do this before recording a mount:
https://github.com/bvaughn/react-devtools-experimental/blob/456e937b83b8c05b0f7cdd6ea0e464018f5d9f19/src/backend/renderer.js#L793-L796

**Edit** Ah, never mind. The `recordUnmount` function does this for us:
https://github.com/bvaughn/react-devtools-experimental/blob/456e937b83b8c05b0f7cdd6ea0e464018f5d9f19/src/backend/renderer.js#L754

I get why these are inconsistent currently, but maybe we should them more consistent at some point to avoid confusion.

PR #181, opened 2019-04-18
Fail tests on errors in renderer

gaearon:
React catches errors in DevTools renderer hooks. However, this means they don't fail tests.
Now they do.

### DIFF
--- a/src/__tests__/setupTests.js
+++ b/src/__tests__/setupTests.js
@@ -16,6 +16,7 @@ env.beforeEach(() => {
   jest.useFakeTimers();
 
   const originalConsoleError = console.error;
+  // $FlowFixMe
   console.error = (...args) => {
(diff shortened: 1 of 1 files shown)


src/__tests__/setupTests.js
```diff
@@ -15,6 +15,15 @@ env.beforeEach(() => {
   // Fake timers let us flush Bridge operations between setup and assertions.
   jest.useFakeTimers();
 
+  const originalConsoleError = console.error;
+  console.error = (...args) => {
```
bvaughn: Might need to `spyOn(console, 'error')` instead?
Screen Shot 2019-04-18 at 9 23 56 AM

gaearon: Can I just flowfixme? I don't know how to use spyOn for this.

bvaughn: Sure lol that's fine too

PR #184, opened 2019-04-18
Harden assertions

gaearon:
- Asserts the node count is the same between reorders
- Asserts we remove children first
- Removes the ID logging which creates a DEV-only side effect

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -239,15 +239,12 @@ export function attach(
         parentFiber == null ? null : getDataForFiber(parentFiber);
       const parentFiberDisplayName =
         (parentFiberData && parentFiberData.displayName) || 'null';
+      // NOTE: calling getFiberID or getPrimaryFiber is unsafe here
+      // because it will put them in the map. For now, we'll omit them.
+      // TODO: better debugging story for this.
       console.log(
-        `[renderer] %c${name} %c${getFiberID(
-          getPrimaryFiber(fiber)
-        )}:${fiberDisplayName} %c${
-          parentFiber
-            ? getFiberID(getPrimaryFiber(parentFiber)) +
-              ':' +
-              parentFiberDisplayName
-            : ''
+        `[renderer] %c${name} %c${fiberDisplayName} %c${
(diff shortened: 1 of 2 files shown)


src/backend/renderer.js
```diff
@@ -239,15 +239,12 @@ export function attach(
         parentFiber == null ? null : getDataForFiber(parentFiber);
       const parentFiberDisplayName =
         (parentFiberData && parentFiberData.displayName) || 'null';
+      // NOTE: calling getFiberID or getPrimaryFiber is unsafe here
+      // because it will put them in the map. For now, we'll omit them.
+      // TODO: better debugging story for this.
```
bvaughn: I can follow up on this.

src/devtools/store.js
```diff
@@ -711,9 +711,17 @@ export default class Store extends EventEmitter {
...
           weightDelta = -element.weight;
 
+          if (element.children.length > 0) {
+            throw new Error(
+              'Fiber ' +
+                id +
+                ' was removed before its children. ' +
+                'This is a bug in React DevTools.'
+            );
+          }
```
bvaughn: Nice.

src/devtools/store.js
```diff
@@ -769,7 +777,16 @@ export default class Store extends EventEmitter {
...
+          const prevChildren = element.children;
           element.children = Array.from(children);
+          if (element.children.length !== prevChildren.length) {
+            throw new Error(
+              'Fiber ' +
+                id +
+                ' received a different number of children on reorder. ' +
+                'This is a bug in React DevTools.'
+            );
+          }
```
bvaughn: I was  adding a version of this locally too that also compared contents to make sure nothing new was added, but I hadn't pushed it yet. I'll just rebase on top of yours! 😁

bvaughn: 24736e4ff57832b9fcc1da8137729ae65a78124a

https://github.com/bvaughn/react-devtools-experimental/blob/60b614f5272fd0fa400c555e011f8262582c5b55/src/devtools/store.js#L776-L787

PR #186, opened 2019-04-19
Fix more Suspense issues

gaearon:
Builds on https://github.com/bvaughn/react-devtools-experimental/pull/184. See individual commits.

The TLDR is that our current approach doesn't guarantee children-first unmount order because in case of Suspense primary->fallback transition, there can be *both* real deletions (such as node getting removed) *and* nodes getting "hidden".

As a fix:

- We now collect unmounted IDs in two separate arrays. One for real unmounts, one for hiding Suspense nodes.
- We track which Fibers _really_ unmounted for a given commit so we don't count them twice.
- The REMOVE operation is now `[REMOVE, length, ...ids]`. It always has many items. We create this operation just before materializing the typed array.
- There's no need for "prepend operation" mode anymore, the operations are always queued at the end.
- New stress tests cover this logic.
- I added block scoping to switches because otherwise they're a pain to work with. The shared weight adjustment block had to be extracted because I need to call it many times for the REMOVE operation now.


src/__tests__/__snapshots__/storeStressSync-test.js.snap
```diff
@@ -115,6 +115,260 @@ exports[`StoreStress (Sync Mode) should handle a stress test for Suspense (Sync
       <Y>
 `;
 
+exports[`StoreStress (Sync Mode) should handle a stress test for Suspense without type change (Sync Mode) 1`] = `
```
bvaughn: Just thinking out loud. I'm not asking for you (or anyone) to do this (particularly for fuzz tests) _but_ I found the `Store` snapshot  tests much easier to read when I used named snapshots. Assuming the names are meaningful, this lets me quickly scan the snapshot file to see if anything jumps out as incorrect. That can be hard to do when they're just numbered.

For example:
https://github.com/bvaughn/react-devtools-experimental/blob/131e2e52db486782efe91867611ab9c7021d27e5/src/__tests__/__snapshots__/store-test.js.snap#L28-L76

src/backend/renderer.js
```diff
@@ -577,13 +574,21 @@ export function attach(
...
+  let pendingRealUnmountedIDs: Array<number> = [];
+  let pendingSimulatedUnmountedIDs: Array<number> = [];
+  let pendingOperationsQueue: Array<Uint32Array> | null = [];
 
+  // We keep track of which Fibers have been reported as unmounted by React
+  // during this commit phase so that we don't try to "hide" them or their
+  // children when Suspense flips to fallback. These Fibers won't have IDs.
+  let fibersUnmountedInThisCommitPhase: WeakSet<Fiber> = new WeakSet();
+
+  // TODO: we could make this layer DEV-only and write directly to pendingOperations.
```
bvaughn: I don't think I understand this comment.

src/backend/renderer.js
```diff
@@ -746,19 +781,22 @@ export function attach(
...
-      // This is why we prepend the delete operation to the queue.
-      endNextOperation(true);
+      // To maintain child-first ordering,
+      // we'll push it into one of these queues,
+      // and later arrange them in the correct order.
+      if (isSimulated) {
+        pendingSimulatedUnmountedIDs.push(id);
+      } else {
+        pendingRealUnmountedIDs.push(id);
+      }
```
bvaughn: Before, this added each entry to the queue in reverse order (via `splice`). Now it no longer reverses the order (since it uses `push`). I'm guessing this is okay or it would fail tests and trigger invariants, but it surprises me. Am I misreading something?

src/devtools/store.js
```diff
@@ -574,25 +601,16 @@ export default class Store extends EventEmitter {
       }
     }
 
-    let addedElementIDs: Uint32Array = new Uint32Array(0);
-    let removedElementIDs: Uint32Array = new Uint32Array(0);
+    let addedElementIDs: Array<number> = [];
+    let removedElementIDs: Array<number> = [];
```
bvaughn: nit:  Can make these `const` now 😄

src/devtools/store.js
```diff
@@ -689,67 +711,72 @@ export default class Store extends EventEmitter {
...
-            addedElementIDs.set(oldAddedElementIDs);
-            addedElementIDs[oldAddedElementIDs.length] = id;
-
-            weightDelta = 1;
+            addedElementIDs.push(id);
+            this._adjustParentTreeWeight(parentElement, 1);
           }
           break;
+        }
         case TREE_OPERATION_REMOVE: {
```
bvaughn: We should update the `OVERVIEW.md` to reflect this change.

src/devtools/store.js
```diff
@@ -768,8 +795,17 @@ export default class Store extends EventEmitter {
             );
           }
 
-          element = ((this._idToElement.get(id): any): Element);
+          const element = ((this._idToElement.get(id): any): Element);
+          const prevChildren = element.children;
           element.children = Array.from(children);
+          if (element.children.length !== prevChildren.length) {
```
bvaughn: Sorry. I hardened this too (to verify that no children were added or removed) and it conflicted.

src/devtools/store.js
```diff
@@ -782,10 +818,13 @@ export default class Store extends EventEmitter {
             });
 
             element.weight = nextWeight;
-
-            weightDelta = nextWeight - prevWeight;
+            // TODO: passing null here is suspicious, but it's existing behavior.
+            // It is suspicious because either we shouldn't need a delta at all--
+            // or we should apply it to all parents, and not just this item.
+            this._adjustParentTreeWeight(null, nextWeight - prevWeight);
```
bvaughn: We should no longer need to call this during a reorder, right? Especially now that we have invariants in place to ensure the contents didn't change, only their order.

Just to be sure, I commented out this whole `if (!element.isCollapsed)` block in both master _and_ your branch, and all of the tests  pass.

src/backend/renderer.js
gaearon: Now that it’s append-only, the intermediate array doesn’t give us anything except extra validation. We could have written into the “operations” instead.

src/backend/renderer.js
gaearon: Maybe you’re missing the part where we’re sending the operations? The React-reported unmounts are added in the reverse order to typed array.

src/devtools/store.js
gaearon: Sure

src/backend/renderer.js
```diff
@@ -616,13 +617,44 @@ export function attach(
...
+
+    // All unmounts except roots are batched in a single message.
+    ops[i++] = TREE_OPERATION_REMOVE;
+    // The first number is how many unmounted IDs we're gonna send.
+    ops[i++] =
+      pendingRealUnmountedIDs.length + pendingSimulatedUnmountedIDs.length;
+    // Fill in the real unmounts in the reverse order.
+    // They were inserted parents-first by React, but we want children-first.
+    // So we traverse our array backwards.
+    for (let j = pendingRealUnmountedIDs.length - 1; j >= 0; j--) {
```
gaearon: This is where we flip the order now.

src/backend/renderer.js
bvaughn: Ah! I missed the fact that you were flipping the order in the other `for` loop. Gotcha!  👍

src/__tests__/__snapshots__/storeStressSync-test.js.snap
gaearon: Sorry, I should've clarified that the idea is you shouldn't need to read these snapshots. Most of the tests are _not_ about them.

These snapshots represent the _initial_ steps. But pretty much all failures I've ran into were around updates. In the updates case, we compare the store content to a snapshot, so the snapshot name itself isn't very relevant or useful. What's relevant is the index of the failing case. I'll write up more about the debugging workflow if you'd like later.

I'm not opposed to giving them names. Just saying I haven't had a single "snapshot failure" while debugging this, it's usually the code below that fails. So I'd rather prefer to add more useful logging there. In either case IMO this shouldn't block this PR.

bvaughn: > In either case IMO this shouldn't block this PR.

I think my original comment made that super clear. I didn't ask you to change anything, _especially_ for fuzz/stress tests. I was just sharing a thought about snapshots in general.

gaearon: Yeah, that totally makes sense in general.

PR #196, opened 2019-04-19
Convert inspected element (right panel) to use Suspense

### LINKED ISSUE #44: Convert inspected element (right panel) to use Suspense

bvaughn:
Resolves #44

### TODO
- [x] Poll renderer interface for updates.
- [x] Schedule an update with React when the inspected element is invalidated.
- [x] Use the "two `setState` pattern" to separate updating the selected element in the tree from suspending. (This does not seem to be working correctly.)
- [x] Avoid re-sending the same data through the bridge if the fiber has not been committed since the last time it was inspected.
- [x] Use an LRU to avoid leaking memory over time as we switch between inspected elements.


bvaughn:
- [x] Refactored `TreeContext` to use less memoization and just export `dispatch` directly.

Regarding this change, I just cherry-picked it into master since I'm not sure how quickly this branch will land given the outstanding jank issue.

bvaughn:
I think all outstanding issues have been fixed now except for the bizarre Flow error message.

I'm going to stake a step away from this PR for a little bit and revisit when I feel like I can take a final pass over the code for anything glaring that I forgot. Feels really nice to use now though!

gaearon:
👏👏👏

package.json
```diff
@@ -128,14 +128,14 @@
     "opener": "^1.5.1",
     "prettier": "^1.16.4",
     "prop-types": "^15.6.2",
-    "react": "0.0.0-4221565e1",
+    "react": "^0.0.0-4221565e1",
```
gaearon: This is suspicious to me. Last time I did that, it gave me a completely different beta version from months ago because it matches alphabetically (its commit hash was later in the alphabet).

**You can verify that `npm install react@^0.0.0-4221565e1` installs `react@0.0.0-fec00a869` — not what you want.** (Although your PR is fine because the Yarn lockfile currently already has the right version.)

Instead, I think what you want is to lock down *both* React packages and the scheduler on the exact canary we want. Without the caret.

src/devtools/views/Components/TreeContext.js
```diff
@@ -712,11 +762,11 @@ function TreeContextController({ children }: Props) {
     store.addListener('mutated', handleStoreMutated);
 
     return () => store.removeListener('mutated', handleStoreMutated);
-  }, [dispatch, initialRevision, store]);
+  }, [dispatchWrapper, initialRevision, store]);
 
   return (
     <TreeStateContext.Provider value={state}>
```
sebmarkbage: Should the lower pri state have its own context so that you don't have to rerender all of those in the high pri render?

package.json
bvaughn: Ah, right that makes sense. This is what I got by running `yarn add react@canary react-dom@canary scheduler@canary`.

src/devtools/views/Components/TreeContext.js
bvaughn: Yeah, probably. An earlier version of this (one of my previous commits) actually had it in a separate context, which used a ref and had its own `unstable_next` call to schedule the derived update. Maybe I can revisit that idea now that things are actually working.

src/devtools/views/Components/SelectedElement.js
```diff
@@ -267,80 +265,3 @@ function OwnerView({ displayName, id }: { displayName: string, id: number }) {
...
-
-    // A null id indicates that there's nothing currently selected in the tree.
-    if (id === null) {
-      return () => {};
-    }
-
-    const rendererID = store.getRendererIDForElement(id);
-
-    // Update the $r variable.
-    bridge.send('selectElement', { id, rendererID });
```
gaearon: Was this intentionally removed? Seems like it wouldn't update `$r` now.

gaearon: (If you intentionally removed it, probably should remove the handler in agent too.)

src/devtools/views/Components/InspectedElementContext.js
```diff
@@ -0,0 +1,176 @@
...
+    const rendererID = store.getRendererIDForElement(selectedElementID);
+
+    let timeoutID: TimeoutID | null = null;
+
+    const sendRequest = () => {
+      timeoutID = null;
+
+      bridge.send('inspectElement', { id: selectedElementID, rendererID });
+    };
+
```
gaearon: Might want to add

```js
bridge.send('selectElement', { id: selectedElementID, rendererID });
```

here to update `$r`.

src/devtools/views/Components/SelectedElement.js
bvaughn: I didn't intentionally remove the $r update. I did intentionally remove the imperative hook. Thanks for pointing out the $r bit.

src/devtools/views/Components/InspectedElementContext.js
bvaughn: 👍

src/devtools/views/Components/TreeContext.js
bvaughn: I hit a snag trying to implement this behavior. I tried moving the `inspectedElementID` into the separate `InspectedElementContext` and updating it at a lower priority whenever the tree context's `selectedElementID` but that didn't quite work. The two approach I considered were:
* A ref-based approach that scheduled the second operation via `unstable_next` during render when the selected ID changed. This didn't work because the update seemed to be getting lumped into the current batch, causing jank.
* A passive effect that scheduled the update. (See diff below for the actual code.) However this didn't work either if I intentionally added delay to the `Agent`'s response. In some cases, trying to change the selected index quickly would sometimes stutter.

```diff
diff --git a/src/devtools/views/Components/InspectedElementContext.js b/src/devtools/views/Components/InspectedElementContext.js
index 1aed0dd..e109f93 100644
--- a/src/devtools/views/Components/InspectedElementContext.js
+++ b/src/devtools/views/Components/InspectedElementContext.js
@@ -7,6 +7,7 @@ import React, {
   useMemo,
   useState,
 } from 'react';
+import { unstable_next as next } from 'scheduler';
 import { createResource } from '../../cache';
 import { BridgeContext, StoreContext } from '../context';
 import { hydrate } from 'src/hydration';
@@ -20,6 +21,7 @@ import type {
 import type { Resource, Thenable } from '../../cache';
 
 type Context = {|
+  inspectedElementID: number | null,
   read(id: number): InspectedElement | null,
 |};
 
@@ -66,6 +68,16 @@ function InspectedElementContextController({ children }: Props) {
   // would itself be blocked by the same render that suspends (waiting for the data).
   const { selectedElementID } = useContext(TreeStateContext);
 
+  // Update the inspected element ID (which may cause us to suspended) at a lower priority.
+  const [inspectedElementID, setInspectedElementID] = useState<number | null>(
+    selectedElementID
+  );
+  useEffect(() => {
+    if (selectedElementID !== inspectedElementID) {
+      next(() => setInspectedElementID(selectedElementID));
+    }
+  }, [inspectedElementID, selectedElementID]);
+
   const [count, setCount] = useState<number>(0);
 
   // This effect handler invalidates the suspense cache and schedules rendering updates with React.
@@ -151,11 +163,12 @@ function InspectedElementContextController({ children }: Props) {
 
   const value = useMemo(
     () => ({
+      inspectedElementID,
       read: resource.read,
     }),
     // Count is used to invalidate the cache and schedule an update with React.
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [count]
+    [count, inspectedElementID]
   );
 
   return (
diff --git a/src/devtools/views/Components/SelectedElement.js b/src/devtools/views/Components/SelectedElement.js
index c2b62a6..e7e0f29 100644
--- a/src/devtools/views/Components/SelectedElement.js
+++ b/src/devtools/views/Components/SelectedElement.js
@@ -23,12 +23,11 @@ import type { Element, InspectedElement } from './types';
 export type Props = {||};
 
 export default function SelectedElement(_: Props) {
-  const { inspectedElementID } = useContext(TreeStateContext);
   const viewElementSource = useContext(ViewElementSourceContext);
   const bridge = useContext(BridgeContext);
   const store = useContext(StoreContext);
 
-  const { read } = useContext(InspectedElementContext);
+  const { inspectedElementID, read } = useContext(InspectedElementContext);
 
   const element =
     inspectedElementID !== null
diff --git a/src/devtools/views/Components/TreeContext.js b/src/devtools/views/Components/TreeContext.js
index 21b16e8..88a9116 100644
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -19,7 +19,6 @@
 
 import React, {
   createContext,
-  useCallback,
   useContext,
   useEffect,
   useLayoutEffect,
@@ -27,11 +26,6 @@ import React, {
   useReducer,
   useRef,
 } from 'react';
-import {
-  unstable_next as next,
-  unstable_runWithPriority as runWithPriority,
-  unstable_UserBlockingPriority as UserBlockingPriority,
-} from 'scheduler';
 import { createRegExp } from '../utils';
 import { BridgeContext, StoreContext } from '../context';
 import Store from '../../store';
@@ -54,9 +48,6 @@ type StateContext = {|
   ownerFlatTree: Array<number> | null,
   ownerStack: Array<number>,
   ownerStackIndex: number | null,
-
-  // Inspection element panel
-  inspectedElementID: number | null,
 |};
 
 type ACTION_GO_TO_NEXT_SEARCH_RESULT = {|
@@ -147,9 +138,6 @@ type State = {|
   ownerStack: Array<number>,
   ownerStackIndex: number | null,
   ownerFlatTree: Array<number> | null,
-
-  // Inspection element panel
-  inspectedElementID: number | null,
 |};
 
 function reduceTreeState(store: Store, state: State, action: Action): State {
@@ -607,24 +595,6 @@ function reduceOwnersState(store: Store, state: State, action: Action): State {
   };
 }
 
-function reduceSuspenseState(
-  store: Store,
-  state: State,
-  action: Action
-): State {
-  const { type } = action;
-  switch (type) {
-    case 'UPDATE_INSPECTED_ELEMENT_ID':
-      return {
-        ...state,
-        inspectedElementID: state.selectedElementID,
-      };
-    default:
-      // React can bailout of no-op updates.
-      return state;
-  }
-}
-
 type Props = {| children: React$Node |};
 
 // TODO Remove TreeContextController wrapper element once global ConsearchText.write API exists.
@@ -657,7 +627,6 @@ function TreeContextController({ children }: Props) {
           state = reduceTreeState(store, state, action);
           state = reduceSearchState(store, state, action);
           state = reduceOwnersState(store, state, action);
-          state = reduceSuspenseState(store, state, action);
 
           // If the selected ID is in a collapsed subtree, reset the selected index to null.
           // We'll know the correct index after the layout effect will toggle the tree,
@@ -696,29 +665,15 @@ function TreeContextController({ children }: Props) {
     ownerStack: [],
     ownerStackIndex: null,
     ownerFlatTree: null,
-
-    // Inspection element panel
-    inspectedElementID: null,
   });
 
-  const dispatchWrapper = useCallback(
-    (action: Action) => {
-      // Run the first update at "user-blocking" priority in case dispatch is called from a non-React event.
-      // In this case, the current (and "next") priorities would both be "normal",
-      // and suspense would potentially block both updates.
-      runWithPriority(UserBlockingPriority, () => dispatch(action));
-      next(() => dispatch({ type: 'UPDATE_INSPECTED_ELEMENT_ID' }));
-    },
-    [dispatch]
-  );
-
   // Listen for host element selections.
   useEffect(() => {
     const handleSelectFiber = (id: number) =>
-      dispatchWrapper({ type: 'SELECT_ELEMENT_BY_ID', payload: id });
+      dispatch({ type: 'SELECT_ELEMENT_BY_ID', payload: id });
     bridge.addListener('selectFiber', handleSelectFiber);
     return () => bridge.removeListener('selectFiber', handleSelectFiber);
-  }, [bridge, dispatchWrapper]);
+  }, [bridge, dispatch]);
 
   // If a newly-selected search result or inspection selection is inside of a collapsed subtree, auto expand it.
   // This needs to be a layout effect to avoid temporarily flashing an incorrect selection.
@@ -742,7 +697,7 @@ function TreeContextController({ children }: Props) {
       addedElementIDs,
       removedElementIDs,
     ]: Array<Uint32Array>) => {
-      dispatchWrapper({
+      dispatch({
         type: 'HANDLE_STORE_MUTATION',
         payload: [addedElementIDs, removedElementIDs],
       });
@@ -753,7 +708,7 @@ function TreeContextController({ children }: Props) {
       // At the moment, we can treat this as a mutation.
       // We don't know which Elements were newly added/removed, but that should be okay in this case.
       // It would only impact the search state, which is unlikely to exist yet at this point.
-      dispatchWrapper({
+      dispatch({
         type: 'HANDLE_STORE_MUTATION',
         payload: [new Uint32Array(0), new Uint32Array(0)],
       });
@@ -762,11 +717,11 @@ function TreeContextController({ children }: Props) {
     store.addListener('mutated', handleStoreMutated);
 
     return () => store.removeListener('mutated', handleStoreMutated);
-  }, [dispatchWrapper, initialRevision, store]);
+  }, [dispatch, initialRevision, store]);
 
   return (
     <TreeStateContext.Provider value={state}>
-      <TreeDispatcherContext.Provider value={dispatchWrapper}>
+      <TreeDispatcherContext.Provider value={dispatch}>
         {children}
       </TreeDispatcherContext.Provider>
     </TreeStateContext.Provider>
```

bvaughn: @acdlite  and I talked, and he felt better about the more idiomatic way of approaching this- scheduling both high and low pri updates at the same time (as the PR currently does)- but when I sat down to implement it, I don't think it makes sense.

The `TreeContext` supports a dozen or so actions. After each action, it's important that the lower priority update is also scheduled. To ensure this happens, I think we'd basically need to pass a `dispatch` wrapper method down (like I'm currently doing).

To make that work with the separate context, the `TreeContext` would need to consume the dispatch method of the `InspectedElementContext`- so it could wrap them. However, the point at which the update is being scheduled with tree, it doesn't know the new `selectedElementID` - so there's nothing meaningful it can schedule with the `InspectedElementContext` (and since the `InspectedElementContext` would need to be above the `TreeContext`, it can't read the value from the `TreeContext` either).

Maybe I could store the tree's current/latest state in a ref and pass it to the `InspectedElementContext` to be parsed, but this feels pretty clunky.

yarn.lock
```diff
@@ -9912,14 +9912,6 @@ scheduler@0.0.0-4221565e1:
...
     object-assign "^4.1.1"
 
-scheduler@^0.14.0-alpha.0:
-  version "0.14.0-alpha.0"
-  resolved "https://registry.yarnpkg.com/scheduler/-/scheduler-0.14.0-alpha.0.tgz#6d301d8fd10373487e0e47e837ad24e863ce807d"
-  integrity sha512-qlVVnhIJqLu9E09ZYYJHG4nYwrS4TG2zfODfS/Mk3oSGjeJlrBZTFSU3yfMdu5lIrYzWmHHWFgERPsKbildS6Q==
-  dependencies:
-    loose-envify "^1.1.0"
-    object-assign "^4.1.1"
-
```
bvaughn: Whoops!

src/devtools/views/hooks.js
```diff
@@ -56,7 +56,7 @@ export function useLocalStorage<T>(
     value => {
       try {
         const valueToStore =
-          value instanceof Function ? value(storedValue) : value;
+          value instanceof Function ? (value: any)(storedValue) : value;
```
bvaughn: The new Flow bin complained about this.

src/devtools/views/Components/types.js
```diff
@@ -65,6 +65,11 @@ export type InspectedElement = {|
   source: Object | null,
 |};
 
+export type InspectedElementResponse = {|
+  id: number,
+  inspectedElement: InspectedElement | null,
```
bvaughn: I'm not sure if this is the most graceful/intuitive way to model this. Essentially we need to distinguish between three possible states:
* No element was found for the specified ID (currently returns `null`).
* An element was found, and here are its props (currently returns a `InspectedElementResponse` with an `InspectedElement` inside).
* The element has not been rendered/committed since the last time it was inspected (currently returns a `InspectedElementResponse` with a null value inside).

Maybe this optimization is premature. I think it might help with cases like the Destiny Item Manager, although I don't think it's sufficient to fix those cases (since hovering over a lot of elements may still cause us to request a lot of inspections initially). So maybe I should just revert this particular change and follow up via #195

src/devtools/cache.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import React, { createContext } from 'react';
+import LRU from 'lru-cache';
```
bvaughn: This complicates the `cache` definition a bit, but I think it's necessary to avoid leaking memory over time as we inspect a lot of elements.

Alternately...maybe I could use a `WeakMap` and use the `Element` object in the `Store` as the key? Thoughts?

src/devtools/ProfilingCache.js
```diff
@@ -293,8 +297,11 @@ export default class ProfilingCache {
     });
 
   invalidate() {
-    // Invalidate Susepnse caches.
-    invalidateResources();
+    // Invalidate Suspense caches.
+    this.CommitDetails.clear();
+    this.FiberCommits.clear();
+    this.Interactions.clear();
+    this.ProfilingSummary.clear();
```
bvaughn: This was over-aggressively invalidating before.

src/devtools/views/Components/types.js
bvaughn: I just reverted this change for now (0aa7d2f). We can re-visit it with an eye to perf for cases like D.I.M.

src/devtools/cache.js
gaearon: Do we promise to never recreate elements? I’m not sure we rely on their identity anywhere. I guess there’s no reason to recreate them..

bvaughn: We create them on mount and otherwise mutate them for e.g. changes in `isCollapsed`, `weight`, or `children`

gaearon: Btw was always curious — is there any reason we recreate children array instead of mutating them on changes?

bvaughn: Yes, but maybe it's a premature optimization 😅 

https://github.com/bvaughn/react-devtools-experimental/blob/8b33dd673d965d23d2516bd30c9c3d05386ce881/src/devtools/store.js#L71-L73

bvaughn: Regarding mutable elements- maybe it's a premature optimization 😅but it seemed worth  doing

https://github.com/bvaughn/react-devtools-experimental/blob/8b33dd673d965d23d2516bd30c9c3d05386ce881/src/devtools/store.js#L71-L73

> is there any reason we recreate children array instead of mutating them on changes?

Honestly? I think laziness.

src/devtools/views/Components/TreeContext.js
bvaughn: I'm going to move forward with this implementation as-is, because I think it's strictly an improvement over the previous  imperative/subscription based approach, but I'm happy to keep talking about ways to refactor this to a more idiomatic form.

src/devtools/cache.js
bvaughn: I went ahead and made this change. Feels nicer.

Issue #189, opened 2019-04-19
In collapsed-by-default mode, Search doesn't scroll to first items


gaearon:
1. Open a large app in collapsed-by-default mode
2. Let it load
3. Start typing "Suspense"

The scroll jumps somewhere but there's no Suspense there. This is because "s" matched something else and then "su" matched Suspense — but it was close enough that it decided not to scroll at all.

It's weird that it decided not to scroll. It should always scroll search result into a viewport. That seems like abug.

gaearon:
![Screen Recording 2019-04-20 at 02 46 PM](https://user-images.githubusercontent.com/810438/56458163-28e90d80-637b-11e9-85b3-281aac5f77dc.gif)

gaearon:
Here's what happens:

* We change the selected index
* That fires the effect in TreeContext which toggles the parent expansion
* Effect in Tree tries to scroll to the node (but it's not expanded yet!)
* Store notifies us of the mutation (with expanded nodes), but it's too late

PR #201, opened 2019-04-20
Fix scrolling to selected row in collapsed mode

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/189.

If we've selected an ID which is inside a collapsed subtree, its index can't be valid. We can't scroll to it. However, we don't want to "forget" the selection either.

Instead, we'll temporarily forget the index (but remember the ID) just before we exit the reducer. When the store is reconciled with selection in the layout effect, the index will be automatically recalculate. That, in turn, will trigger the scroll effect to the right index.

### DIFF
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -596,6 +596,20 @@ function TreeContextController({ children, viewElementSource }: Props) {
           state = reduceTreeState(store, state, action);
           state = reduceSearchState(store, state, action);
           state = reduceOwnersState(store, state, action);
+
+          // If the selected ID is in a collapsed subtree, reset the selected index to null.
+          // We'll know the correct index after the layout effect will toggle the tree,
+          // and the store tree is mutated to account for that.
+          if (
+            state.selectedElementID !== null &&
+            store.isInsideCollapsedSubTree(state.selectedElementID)
+          ) {
+            return {
+              ...state,
+              selectedElementIndex: null,
+            };
+          }
+
           return state;
         default:
           throw new Error(`Unrecognized action "${type}"`);
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -429,6 +429,21 @@ export default class Store extends EventEmitter {
     return null;
   }
 
+  isInsideCollapsedSubTree(id: number): boolean {
+    let current = this._idToElement.get(id);
+    while (current != null) {
+      if (current.parentID === 0) {
+        return false;
+      } else {
+        current = this._idToElement.get(current.parentID);
+        if (current != null && current.isCollapsed) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
   startProfiling(): void {
     this._bridge.send('startProfiling');
 


src/devtools/views/Components/TreeContext.js
```diff
@@ -596,6 +596,20 @@ function TreeContextController({ children, viewElementSource }: Props) {
...
           state = reduceSearchState(store, state, action);
           state = reduceOwnersState(store, state, action);
+
+          // If the selected ID is in a collapsed subtree, reset the selected index to null.
+          // We'll know the correct index after the layout effect will toggle the tree,
+          // and the store tree is mutated to account for that.
+          if (
+            state.selectedElementID !== null &&
+            store.isInsideCollapsedSubTree(state.selectedElementID)
+          ) {
```
bvaughn: This makes me nervous.

gaearon: Not sure what you mean specifically

bvaughn: This change makes me nervous- temporarily setting an incorrect state and relying on an effect's cascading update to fix it up for us. Also because of the potential for impact with the two `setState` stuff I'm working on over in PR #196.

gaearon: Hmm. I wouldn't say it's "incorrect".

I'd say the state is "incorrect" when index refers to a different item than is selected. Which is what used to happen.

In this case we're explicitly acknowledging that `selectIndex` is *unknown*. It's only incorrect if other code can't deal with that — but afaik it does.

gaearon: Here's another concrete example where modeling unknown value explicitly makes a difference.

Before this fix, if you selected a node, collapsed its parent, and pressed "down", the selection would "jump" way below: 

Screen Recording 2019-04-22 at 10 26 PM

This is because it would still be using the index — and the index is not valid anymore! So we end up in a difference place in the tree.

With this fix, pressing "down" when selected node is collapsed always navigates you to the root element.

I'm not saying this is necessarily a better behavior (maybe it should navigate to the closest non-collapsed parent, or maybe it should expand instead). But I'm saying that this fix exposes an existing problem which doesn't go away even if we didn't merge it.

bvaughn: Hm. I think that sounds like a case where our `HANDLE_STORE_MUTATION` action just does the wrong thing. Rather than checking if the item is still in the tree, it should check to see if it's still in the *visible* tree, and if not, bump selection up to its nearest visible ancestor.

gaearon: That would be great!

gaearon: However that also reveals a broader problem. If you a node whose child is selected gets deleted, we currently completely lose the selection. However instead we should select the closest non-deleted (and non-collapsed while we’re at it) parent. Or sibling. But definitely not lose the selection like we do now.

bvaughn: > If a node whose child is selected gets deleted, we currently completely lose the selection.

I don't think I understand. The child would also be _explicitly_ deleted in such a case, right?

gaearon: Sorry, I misspoke. I mean this:

Select node
Delete that node

Ideal behavior: selection either goes to sibling or to parent.

Current behavior: selection disappears.

I wanted to fix this by selecting parent. But by the time node is removed we can’t query store about its parent anymore. I guess we could fix this by making removedIds a map that suggests next selection to the tree context mutation handler.

bvaughn: When you say  "current behavior"  do you mean prior to this change landing? or current-current?

PR #204, opened 2019-04-20
Replace WeakSet with return pointer check

gaearon:
There is a simpler (and likely more efficient) way to check if a Fiber has really been unmounted — its `.return` pointer would be nulled out. I was feeling uneasy due to that `WeakSet` because I'm worried putting Fibers into weak sets can mess with VM optimizations. Such as if they turn it into an expando internally. This avoids it.

Curiously, technically even the `.return` check isn't necessary here anymore. If the Fiber is deleted, it wouldn't be in the map — and so `recordUnmount` would have bailed out by now because of its inner check. The only reason this didn't work before was because in `__DEV__` we used to call `getFiberID` for them in `debug()` calls, making those Fibers go into the map. However, even though this is fixed, I don't want to rely on this, and would prefer those `.return` guards to be explicit.

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -589,11 +589,6 @@ export function attach(
   let pendingSimulatedUnmountedIDs: Array<number> = [];
   let pendingOperationsQueue: Array<Uint32Array> | null = [];
 
-  // We keep track of which Fibers have been reported as unmounted by React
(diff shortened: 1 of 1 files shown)


bvaughn:
> The only reason this didn't work before was because in __DEV__ we used to call getFiberID for them in debug() calls, making those Fibers go into the map.

D'oh!

PR #205, opened 2019-04-20
[BLOCKED] Use new @reach/tooltip

bvaughn:
Based on early user testing, I think we need to move away from native tooltips. Ryan recently added tooltips to Reach UI and they seem promising.

I was also able to delete our local fork of Portal since Ryan merged and released https://github.com/reach/reach-ui/pull/137.

### Before
Native tooltips take too long to show up in the browser:
tooltip-before-Kapture 2019-04-20 at 13 26 15

### After
Reach UI tooltips are much faster:
tooltip-after-Kapture 2019-04-20 at 13 45 14

Transitions seem smooth too:
tooltips-Kapture 2019-04-20 at 13 59 29

They also show on tab focus, which seems like a nice touch:
light-Kapture 2019-04-20 at 14 20 17

# Bugs
Unfortunately, sometimes tooltips get orphaned, for example:
bug-Kapture 2019-04-20 at 14 08 26

Also it looks like the `MenuButton` component (another Reach UI component we use) doesn't properly work with the new `Tooltip` component yet, since it only supports function refs and the new `Tooltip` uses a ref hook.


bvaughn:
The colors aren't quite right, but this is getting closer:
tab-key-focus-aahh-Kapture 2019-04-20 at 15 40 53

I'm changing the "focused" styles to only show the outline when the focus comes from a tab key (and not from a click).

bvaughn:
Here's what the owners list and drop-down look like with the new visual treatment:
owners-dropdown-Kapture 2019-04-21 at 7 57 54

owners-focus-Kapture 2019-04-21 at 7 57 03

bvaughn:
This PR has turned into a largish refactoring of hover+focus styles based on some user feedback.  I'm  going to disable tooltips for now (and leave a follow up issue #209 to re-enable them once the bug is fixed) and land this PR because of the contrast, color, and keyboard-focus changes.

Issue #131, opened 2019-04-10
Bounding boxes don't include all children


sophiebits:
Repro:

1. Hover EditableProps on https://react-devtools-experimental.now.sh/

Expected: Everything from the "Editable props" header through the list under "Forward Ref" is highlighted
Actual: Only the "Editable props" header is highlighted

I'm guessing it is only finding the first DOM node in the hovered component instead of all of them.

gaearon:
Yeah, I think old DevTools has the same issue. Would be nice to fix.

sophiebits:
Diagnosis:

This happens because highlightElementInDOM calls into findNativeByFiberID, which ultimately calls into React's findHostInstanceByFiber which only returns the first (like findDOMNode does). (findNativeByFiberID has some weird Suspense logic too, which I don't think belongs in devtools: https://github.com/bvaughn/react-devtools-experimental/pull/61/files#r274755172.)

To fix, I'd probably add another findHostInstance**s**ByFiber (plural) method in React's injectIntoDevTools that returns an array.

sompylasar:
Sneak peek:
<img width="817" alt="Screen Shot 2019-04-20 at 11 48 02 PM" src="https://user-images.githubusercontent.com/498274/56466415-d3d2e900-63c6-11e9-9979-c475bb7d2716.png">

sompylasar:
Also (can be in a separate PR but related) direction hint in the tooltip to see where to look for your elements:
<img width="818" alt="Screen Shot 2019-04-20 at 11 49 44 PM" src="https://user-images.githubusercontent.com/498274/56466437-27ddcd80-63c7-11e9-8d85-e75097032805.png">

sompylasar:
PRs submitted:
- multiple highlights: https://github.com/bvaughn/react-devtools-experimental/pull/207
- arrow direction in the tooltip (builds on top of the highlights rework): https://github.com/bvaughn/react-devtools-experimental/pull/208

sompylasar:
Thanks @sophiebits for the investigation and the suggested solution!

PR #207, opened 2019-04-21
Support highlighting of all DOM elements of Fragments, not just first

sompylasar:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/131

Uses the new function `findAllCurrentHostFibers`.

Removes dependency on React's `renderer.findHostInstanceByFiber` function
which used to highlight only the first DOM element of a Fragment.

Reworked `src/backend/views/Overlay` to support highlighting
more than one element rectangle annotated with one tooltip.
Fixed minor issues with the tooltip position calculation.

Screen Shot 2019-04-20 at 11 48 02 PM

### DIFF
--- a/src/backend/agent.js
+++ b/src/backend/agent.js
@@ -276,18 +276,19 @@ export default class Agent extends EventEmitter {
       console.warn(`Invalid renderer id "${rendererID}" for element "${id}"`);
     }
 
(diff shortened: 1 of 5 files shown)


gaearon:
Lol I was just looking at this PR :P

bvaughn:
Whoops sorry 😄

PR #208, opened 2019-04-21
Add arrow to overlay tip pointing where to look for offscreen elements

sompylasar:
This change builds on top of the overlay rework for multiple highlighted rectangles:
https://github.com/bvaughn/react-devtools-experimental/pull/207

If the highlighted elements are offscreen, the tooltip is positioned
towards these elements, but the element highlights (OverlayRect)
are not visible, so if the element size is zero or close to it,
it's hard to visually tell apart an onscreen element located where
the tooltip is positioned and an offscreen element located elsewhere.

This change improves the UX by adding to the tooltips
a little arrow pointing to where the elements are located
if they are offscreen.

For onscreen elements the tooltip does not change.

overlay-tip-direction-arrow-demo

### DIFF
--- a/shells/dev/app/Offscreen/index.js
+++ b/shells/dev/app/Offscreen/index.js
@@ -1,0 +1,50 @@
+import React from 'react';
+
+function Positioned({ style }) {
+  return (
+    <div style={{ position: 'absolute', width: 0, height: 0, ...style }} />
+  );
+}
+
+function OffscreenLeftTop() {
+  return <Positioned style={{ left: '-200vw', top: '-200vh' }} />;
+}
+function OffscreenTop() {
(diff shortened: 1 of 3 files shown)


sompylasar:
Updated to use the shape of the tooltip to point to elements instead of a unicode character in the tooltip text, please take a look.

The code may be harder to read due to trigonometry but the idea is to show diagonal directions, too, not just top-right-bottom-left.

overlay-tip-direction-arrow-demo

bvaughn:
Closing all PRs at this repo as part of migrating to github.com/facebook/react

sompylasar:
Do you still want this, PRed to the react repo?

PR #210, opened 2019-04-21
Implement showing native elements under a persisted Settings item

sompylasar:
It's often necessary to debug native (host) elements rendered by
the components, but the new version of DevTools removed this functionality.

This change restores the functionality under a persisted setting,
working similar to "reload and profile", so that by default
the performance of the DevTools is unaffected.

To distinguish the HostComponent and HostText types in Components,
new ElementType items were added, and new CSS variables were added
for colors in both light and dart themes.

Within this diff there's also a fix to the shells/dev iframe reload handling.
Previously reloading the iframe would not restore the test app
and would break the bridge/store. This diff more accurately resembles
the load/unload/messaging behavior and the code written for the extension.

The reasoning for this feature to exist:

> I want to do the reverse: from the tree find where are my elements.
> Not every element that gets rendered is visible and of clickable size.
> https://twitter.com/sompylasar/status/1118580156430311424
> https://twitter.com/dan_abramov/status/1118486315736158208

Kapture 2019-04-21 at 15 11 14

CC @gaearon

### DIFF
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -68,6 +68,9 @@ function createPanelIfReactLoaded() {
           localStorage.setItem(LOCAL_STORAGE_SUPPORTS_PROFILING_KEY, 'true');
           chrome.devtools.inspectedWindow.eval('window.location.reload();');
         });
+        bridge.addListener('reloadAppForShowNativeElements', () => {
+          chrome.devtools.inspectedWindow.eval('window.location.reload();');
+        });
         bridge.addListener('exportFile', ({ contents, filename }) => {
           chrome.runtime.sendMessage({
             exportFile: true,
@@ -109,6 +112,7 @@ function createPanelIfReactLoaded() {
           supportsFileDownloads: browserName === 'Chrome',
           supportsReloadAndProfile: true,
           supportsProfiling,
+          supportsShowingNativeElements: true,
         });
 
         // Initialize the backend only once the Store has been initialized.
--- a/shells/dev/src/devtools.js
+++ b/shells/dev/src/devtools.js
@@ -2,7 +2,7 @@
 
 import { createElement } from 'react';
 // $FlowFixMe Flow does not yet know about createRoot()
-import { unstable_createRoot as createRoot } from 'react-dom';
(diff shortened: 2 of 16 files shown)


bvaughn:
Hey @sompylasar. Thanks for the PR and the feedback about this feature. Based on the feedback that you and a few others have shared, I've been convinced that there's value in providing an option to show host nodes. 😄

That being said, I'd like to approach the feature of filtering with a few broader goals in mind (rather than tackling just the host nodes preference). Those goals include:
* Filtering preferences need to support both temporary values (e.g. things you do while debugging- like showing only `Suspense` nodes) and persisted values (e.g. things you have a subject preference for- like always showing host nodes).
* Preferences will require a format that can be shared between the frontend and backend. (The backend will need to be able to read the preference values during startup, so that it can properly filter the tree, before the frontend has even been rendered.)
* Filtering should be powerful enough to also eventually support user-specified, custom filters (e.g. hide all components that start with a certain name, or within a specific folder on the file system).
* Filtering changes should be able to be applied without requiring a reload of the app. (Toggling the `Suspense`-only mode should be doable without losing your app's current state.) For this to work, we'll need to clear all of the `Store`'s current nodes (and notify contexts like `TreeContext`), then request the backend to re-crawl and initialize the tree with the new filter criteria in mind.

The filtering features above are something I've been considering for a while (although admittedly, I didn't initially plan to include the host nodes preference as one of them). I think they're also a blocker for some other features like Dan's suspense toggling prototype (#112). Because of this, I'm going to work on filtering for the next few days. I'll keep you posted.

sompylasar:
Thanks for the writeup @bvaughn! This strategy is very structured and promising. My only concern is it would take quite some time to implement in full while there could be a quick win now for just a couple but important cases.

bvaughn:
I think it's something I could implement in a few days (hopefully finishing up sometime next week, depending on what else comes up). Timing is important since it also blocks our advanced suspense integration.

sompylasar:
Okay let's see, I'm down 💯 if it takes no more than a few days, and I'll happily update this branch as soon as some draft of the filtering preferences appears. Thank you!

bvaughn:
This has been superseded by #245 so I'm just closing it as cleanup.

PR #223, opened 2019-04-24
Owners list view should drill through intermediate components

### LINKED ISSUE #222: Owners list doesn't drill through intermediate components
Repro: https://codesandbox.io/s/10908p3y27

1. Select `<Intermedia>` and `<Leaf>` nodes in the tree. Both show "rendered by `<Root>`.
2. Double click `<Root>` to see its owner list. It only shows `<Intermedia>` (and not `<Leaf>`).

This is because we stop recursing when we reach a node that isn't owned by the target:
https://github.com/bvaughn/react-devtools-experimental/blob/942f67c9369cc587708bd0cde1e61099ab9ed8f3/src/devtools/views/Components/TreeContext.js#L799-L805

But this doesn't account for intermediate components. I think the proper solution would be to either recurse all the way _or_ maintain an owners map in the `Store` so that we could quickly get a list of all components owned by a target element.

We'll also have to re-think our indentation calculation, since drilling through intermediates would cause the tree to have a funky indentation in its current form.

bvaughn:
Resolves #222

- [x] Moved owners list calculations into the `Store`.
- [x] Added snapshot tests.
- [x] Fixed original indirection bug.
- [x] Added an owners `Map` to the store so we don't have to recurse the full subtree to calculate the list.

### DIFF
--- a/src/__tests__/__snapshots__/storeOwners-test.js.snap
+++ b/src/__tests__/__snapshots__/storeOwners-test.js.snap
@@ -1,0 +1,20 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Store owners list should drill through intermediate components: 1: mount 1`] = `
+[root]
+  ▾ <Root>
+    ▾ <Intermediate>
+      ▾ <Wrapper>
+          <Leaf key="children">
+`;
+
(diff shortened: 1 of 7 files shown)


bvaughn:
This feels like a big improvement on internal FB sites I tested on, so I'm going to move forward with it before my next round of user testing this afternoon. Will forward-fix bugs or implement any suggested changes!

PR #226, opened 2019-04-25
Improve performance by mutating the children array

gaearon:
I'm seeing ~15% improvement for additions in ~20% improvement for deletions on huge trees with this change. Probably because we copy less and because we don't churn the GC as much.

### DIFF
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -720,7 +720,7 @@ export default class Store extends EventEmitter {
             const parentElement = ((this._idToElement.get(
               parentID
             ): any): Element);
-            parentElement.children = parentElement.children.concat(id);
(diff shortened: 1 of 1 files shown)


bvaughn:
> I'm seeing ~15% improvement for additions in ~20% improvement for deletions on huge trees with this change.

Neat. What are the absolute numbers?

gaearon:
With thousands of nodes, I was seeing something like 85ms -> 73ms and 32ms -> 25ms for time spent in add and remove, respectively.

bvaughn:
Nice!

gaearon:
That said in general GC churn accumulates and causes pauses even if each individual operation doesn't take that long.

PR #227, opened 2019-04-25
Transfer strings in a string table

gaearon:
When a large tree mounts or updates, in most cases strings are repetitive. For example, when interacting with a real complex app, I'm seeing how in every batch of `operations`, only ~10% of displayNames/keys are unique. The other ~90% are the same names that we encode multiple times. In the past, we've optimized encoding them to be cached. However, decoding is still lazy. It turns out that **decoding repetitive strings accounts for 40% of time spend in Store's `operations` handler**.

This fixes it with a string table. Individual events like `ADD` no longer contain the encoded strings. Instead, each batch of `operations` contains a list of unique strings in the beginning, and then further operations mention them by ID. I updated the `OVERVIEW` to explain the new format.

## Synthetic results

**Before:** 40ms in renderer, 61ms in the store
**After:** 30ms in renderer, 34ms in the store

## Product code results

**Before:** 109ms in renderer, 246ms in the store
**After:** 108ms in renderer, 150ms in the store

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -590,6 +590,9 @@ export function attach(
   let pendingRealUnmountedIDs: Array<number> = [];
   let pendingSimulatedUnmountedIDs: Array<number> = [];
   let pendingOperationsQueue: Array<Uint32Array> | null = [];
+  let pendingStringTable: Map<string, number> = new Map();
+  let pendingStringTableLength = 0;
+  let pendingStringTableCounter = 0;
 
   function pushOperation(op: number): void {
     if (__DEV__) {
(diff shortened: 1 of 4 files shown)


src/backend/renderer.js
```diff
@@ -672,6 +688,26 @@ export function attach(
...
+
+  function getStringID(str: string | null): number {
+    if (str === null) {
+      return 0;
+    }
+    const existingID = pendingStringTable.get(str);
+    if (existingID !== undefined) {
+      return existingID;
+    }
+    let id = ++pendingStringTableCounter;
```
bvaughn: Why do we need a separate `pendingStringTableCounter` variable instead of just using `pendingStringTable.size`?

gaearon: Good catch. Fixed.

gaearon: Hmm. Let me think more about this for a minute.

gaearon: Yeah that should be good.

Issue #217, opened 2019-04-23
Components tree is sometimes unexpectedly empty after navigation


gaearon:
1. Open FB page
2. Open Components tab
3. Change address bar to `https://reactjs.org` and press Enter

Expected: Components tab gets populated.
Actual:

<img width="783" alt="Screen Shot 2019-04-23 at 7 27 37 PM" src="https://user-images.githubusercontent.com/810438/56606380-00247a80-65fe-11e9-988c-2ad3e69eb579.png">

~~If I **inspect background page**, I see this:~~ (fixed by #229)

<img width="652" alt="Screen Shot 2019-04-23 at 7 27 12 PM" src="https://user-images.githubusercontent.com/810438/56606408-0b77a600-65fe-11e9-9f65-5502401b7e4a.png">

sompylasar:
Please see https://github.com/bvaughn/react-devtools-experimental/pull/229

bvaughn:
I don't think #229 completely resolves this issue. We no longer error, which is good. But the Components tree is usually empty after navigating to reactjs.org.

sompylasar:
> But the Components tree is usually empty after navigating to reactjs.org.

Wow, would you please share a repro, maybe a screen recording? I didn't see that.

bvaughn:
While working on #235 I observed that this behavior is pretty easy to reproduce after going to an internal FB SSR site and then navigating to react-window.now.sh. Doesn't happen _every_ time but happens _most_ times that I need to reload the page after navigating in this case.

bvaughn:
I've observed that I can reliably navigate between [react-window.now.sh](https://react-window.now.sh) and [bvaughn.github.io/react-virtualized](http://bvaughn.github.io/react-virtualized) and the Components tab is never empty. Seems like I can add [reactjs.org](https://reactjs.org) into that mix too and things work okay.

But when I add Facebook (FB5) into the mix and then navigate to [react-window.now.sh](https://react-window.now.sh), the tree is pretty reliably empty.

bvaughn:
Looks like the comment in this line is not accurate:
https://github.com/bvaughn/react-devtools-experimental/blob/34f6fa4fab502497b6ffd5d04cf0547bb9569fc2/shells/browser/shared/src/main.js#L66-L70

sompylasar:
I've also seen some warnings "shutdown" on a destroyed bridge.

There are most likely race conditions between navigation events, messages, and port destruction.

[...]

PR #229, opened 2019-04-26
Fix for 'Attempting to use a disconnected port object'

sompylasar:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/217

The error reproduces with any two React websites, e.g. `https://reactjs.org` and `https://nextjs.org`, by keeping the DevTools Components tab open and switching between these websites in the same browser tab.

There are several issues with the code that contribute to this:
1. `Bridge` leaves behind a dangling timer that fires `_flush` after the bridge has been abandoned ("shutdown").
2. `bridge.send('shutdown')` is asynchronous, so the event handlers do not get unsubscribed in time.
3. `port.onDisconnect` does not trigger on in-tab navigation like new URL or back/forward navigation.
4. State management design of the code that uses shared variables and callbacks makes it hard to handle race conditions originating from the browser.

This commit cleans up some of the lacking symmetry when using `addListener`/`removeListener`, but the code in `shells/browser/shared/src/main.js` is hard to reason about with regards to race conditions, and there are many possible race conditions originating from the browser, so maybe there could be a better design paradigm (like a formal state machine) to manage the state changes in response to sequences of events than plain old event listeners, callbacks, and shared variables.

Unrelated, but clicking Chrome Back/Forward/Back/Forward very fast makes the browser and the DevTools and the DevTools of DevTools stall and become unresponsive for some time, then recovers but the Back/Forward/Stop/Refresh button and favicon loading indicator may remain broken. Looks like a Chrome bug, some kind of a temporary deadlock in handling the browser history.

### DIFF
--- a/shells/browser/shared/src/main.js
+++ b/shells/browser/shared/src/main.js
@@ -46,22 +46,24 @@ function createPanelIfReactLoaded() {
       let root = null;
 
       function initBridgeAndStore() {
-        let hasPortBeenDisconnected = false;
         const port = chrome.runtime.connect({
           name: '' + chrome.devtools.inspectedWindow.tabId,
         });
-        port.onDisconnect.addListener(() => {
-          hasPortBeenDisconnected = true;
-        });
+        // Looks like `port.onDisconnect` does not trigger on in-tab navigation like new URL or back/forward navigation,
+        // so it makes no sense to handle it here.
 
         bridge = new Bridge({
           listen(fn) {
-            port.onMessage.addListener(message => fn(message));
+            const listener = message => fn(message);
+            // Store the reference so that we unsubscribe from the same object.
+            const portOnMessage = port.onMessage;
+            portOnMessage.addListener(listener);
+            return () => {
+              portOnMessage.removeListener(listener);
+            };
           },
           send(event: string, payload: any, transferable?: Array<any>) {
-            if (!hasPortBeenDisconnected) {
-              port.postMessage({ event, payload }, transferable);
-            }
+            port.postMessage({ event, payload }, transferable);
           },
         });
         bridge.addListener('reloadAppForProfiling', () => {
@@ -270,7 +272,8 @@ function createPanelIfReactLoaded() {
 
       // Shutdown bridge and re-initialize DevTools panel when a new page is loaded.
       chrome.devtools.network.onNavigated.addListener(function onNavigated() {
-        bridge.send('shutdown');
+        // `bridge.shutdown()` will remove all listeners we added, so we don't have to.
+        bridge.shutdown();
 
         // It's easiest to recreate the DevTools panel (to clean up potential stale state).
         // We can revisit this in the future as a small optimization.
--- a/src/bridge.js
+++ b/src/bridge.js
@@ -14,19 +14,25 @@ type Message = {|
 export default class Bridge extends EventEmitter {
   _messageQueue: Array<any> = [];
   _timeoutID: TimeoutID | null = null;
+  _destroyed: boolean = false;
   _wall: Wall;
+  _wallUnlisten: Function | null = null;
 
   constructor(wall: Wall) {
     super();
 
     this._wall = wall;
 
-    wall.listen((message: Message) => {
+    this._wallUnlisten = wall.listen((message: Message) => {
       this.emit(message.event, message.payload);
     });
   }
 
   send(event: string, payload: any, transferable?: Array<any>) {
+    if (this._destroyed) {
+      return;
+    }
+
     // When we receive a message:
     // - we add it to our queue of messages to be sent
     // - if there hasn't been a message recently, we set a timer for 0 ms in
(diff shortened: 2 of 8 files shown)


shells/browser/shared/src/main.js
```diff
@@ -45,37 +45,80 @@ function createPanelIfReactLoaded() {
...
+      // By having a bridgeId we ensure that we stop processing the old bridge right before the new bridge is created
+      // to prevent the old bridge from sending and receiving the messages meant for the newly created one.
+      let bridgeId = null;
+
       function initBridgeAndStore() {
-        let hasPortBeenDisconnected = false;
+        // Creating a bridge identifier separately from the `Bridge` instance and before that because
+        // the `listen` passed into `Bridge` constructor is called synchronously within this constructor,
+        // so we have no opportunity to assign the reference to the instance to a variable before this happens.
+        const thisBridgeId = (bridgeId = Math.random());
```
gaearon: Why not an autoincrementing counter? Might make it easier to log/debug too.

sompylasar: Can be a counter. Will need a third variable to store the counter state. I tried it initially but it looked too much so I changed to random that holds this third variable implicitly.

gaearon: Just `let nextId = 0` and using `nextId++` seems fine to me.

gaearon:
Instead of working around this on the consuming side of `Bridge`, can we add a `shutdown` method to bridge that would:

* send `shutdown` event
* cancel the dangling timer
* remove all registered listeners from itself

Then it wouldn't fire old events and we wouldn't need an ID.

sompylasar:
> Instead of working around this on the consuming side of `Bridge`, can we add a `shutdown` method to bridge that would:
> 
> * send `shutdown` event
> * cancel the dangling timer
> * remove all registered listeners from itself
> 
> Then it wouldn't fire old events and we wouldn't need an ID.

@gaearon You won't believe me but this was my initial solution (but I called it `destroy` as I usually do, not `shutdown`). Then I thought that I'm not sure that the `'shutdown'` event is supposed to be the final event, so I changed it to only "unsubscribe" from the bridge, but not actually change its own state to prevent it from sending events to other subscribers (except for synchronously flushing the events which doesn't change its state, just forces it to act quicker than it otherwise would).

gaearon:
I think it is supposed to be the final event.

sompylasar:
Updated, please take a look. Thanks!

shells/browser/shared/src/main.js
```diff
@@ -46,22 +46,24 @@ function createPanelIfReactLoaded() {
...
       function initBridgeAndStore() {
-        let hasPortBeenDisconnected = false;
         const port = chrome.runtime.connect({
           name: '' + chrome.devtools.inspectedWindow.tabId,
         });
-        port.onDisconnect.addListener(() => {
-          hasPortBeenDisconnected = true;
-        });
+        // Looks like `port.onDisconnect` does not trigger on in-tab navigation like new URL or back/forward navigation,
+        // so it makes no sense to handle it here.
```
bvaughn: Hm. I wonder if this used to be true but changed... I think this bit of code was copied verbatim from old DevTools

src/bridge.js
```diff
@@ -14,19 +14,25 @@ type Message = {|
 export default class Bridge extends EventEmitter {
   _messageQueue: Array<any> = [];
   _timeoutID: TimeoutID | null = null;
+  _destroyed: boolean = false;
   _wall: Wall;
+  _wallUnlisten: Function | null = null;
```
bvaughn: I think you added a `null` type here in case a Wall doesn't return an unsubscribe function? But in that case, wouldn't the value actually be `undefined`?

It also looks like maybe all of the listen functions return unlisten functions after this PR? (And you updated the `Wall` type to require this too.)

src/bridge.js
```diff
@@ -41,7 +47,47 @@ export default class Bridge extends EventEmitter {
     }
   }
 
+  shutdown() {
+    if (this._destroyed) {
```
bvaughn: Should we log a warning here?

src/bridge.js
```diff
@@ -14,19 +14,25 @@ type Message = {|
 export default class Bridge extends EventEmitter {
   _messageQueue: Array<any> = [];
   _timeoutID: TimeoutID | null = null;
+  _destroyed: boolean = false;
```
bvaughn: subjective nit: Not a fan of the name `_destroyed`. I think `isShutdown` would be a bit more consistent with other similar var names within the repo.

src/bridge.js
sompylasar: Oops this looks like a leftover from the version where the absence/presense of a wall or unsubscribe function described that the bridge is destroyed/shutdown or not. Now there's a separate flag for that.

src/bridge.js
sompylasar: Yes, leftover from when the shutdown function was called destroy. It's become a muscle memory for me to write these destroy/_destroyed patterns.

src/bridge.js
sompylasar: Yes, would be nice.

PR #230, opened 2019-04-26
Assert empty maps when there are no roots

gaearon:
I thought this might be a useful check for tests. This catches if we forget to delete something from a map in a store.


sompylasar:
@bvaughn I've just hit `Expected _profilingOperations to be empty` when I wrote a test that involves `startProfiling`. Looks like I'll have to fix https://github.com/bvaughn/react-devtools-experimental/issues/232 to unblock.

PR #239, opened 2019-04-28
Remove Agent addBridge: there can only be one bridge, add in constructor

sompylasar:
Let's make impossible states truly impossible, and fix Flow types, too.

All three usages of Agent called addBridge right after constructing it.
Agent has one field `_bridge` which is force-typed as not-null despite
there's a temporary zone between the constructor end and addBridge start
where `_bridge` is null.


bvaughn:
Nice!

Issue #165, opened 2019-04-16
Add Flow types for Fiber


gaearon:
This is effectively as good as `any`:

https://github.com/bvaughn/react-devtools-experimental/blob/20193f03c04971a14ee8a0b6755bdfee08b94ba3/src/backend/types.js#L10-L11

As a result we're missing a lot of coverage in the renderer.

This can be a starting point:

```js
export type Fiber = {|
  tag: number,
  key: null | string,
  type: any,
  stateNode: any,

  return: Fiber | null,
  child: Fiber | null,
  sibling: Fiber | null,
  index: number,

  pendingProps: any,
  memoizedProps: any,
  memoizedState: any,

  effectTag: number,

  alternate: Fiber | null,

  actualDuration?: number,
  actualStartTime?: number,
  selfBaseDuration?: number,
  treeBaseDuration?: number,

  // __DEV__ only
  _debugOwner?: Fiber | null,
|};
```

 It currently creates a lot of violations.

PR #240, opened 2019-04-28
Add Flow types for Fiber

sompylasar:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/165

CC @gaearon

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -26,6 +26,7 @@ shells/dev/build/*
 
 [options]
 esproposal.class_instance_fields=enable
+esproposal.optional_chaining=enable
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe
 suppress_comment=\\(.\\|\n\\)*\\$FlowIssue
 suppress_comment=\\(.\\|\n\\)*\\$FlowIgnore
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -573,7 +573,6 @@ export function attach(
     }
   }
 
-  // eslint-disable-next-line no-unused-vars
   function haveProfilerTimesChanged(
     prevFiber: Fiber,
     nextFiber: Fiber
@@ -715,7 +714,7 @@ export function attach(
     const isProfilingSupported = fiber.hasOwnProperty('treeBaseDuration');
     if (isProfilingSupported) {
       idToRootMap.set(id, currentRootID);
-      idToTreeBaseDurationMap.set(id, fiber.treeBaseDuration);
+      idToTreeBaseDurationMap.set(id, fiber.treeBaseDuration || 0);
     }
 
     const hasOwnerMetadata = fiber.hasOwnProperty('_debugOwner');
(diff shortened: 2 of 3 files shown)


src/backend/types.js
```diff
@@ -7,8 +7,88 @@ type BundleType =
...
+  effectTag: SideEffectTag,
+
+  nextEffect: Fiber | null,
+
+  firstEffect: Fiber | null,
+  lastEffect: Fiber | null,
+
+  expirationTime: ExpirationTime,
+
+  childExpirationTime: ExpirationTime,
```
gaearon: We should omit the properties we don't use in DevTools.

sompylasar: > We should omit the properties we don't use in DevTools.

👉 https://github.com/bvaughn/react-devtools-experimental/pull/240/commits/d35cf80921245ff7da4bcc7c17f68082f21088fb

src/backend/renderer.js
```diff
@@ -1550,6 +1580,11 @@ export function attach(
     // Find the currently mounted version of this fiber (so we don't show the wrong props and state).
     fiber = findCurrentFiberUsingSlowPath(fiber);
 
+    if (fiber == null) {
+      console.warn(`Could not find Fiber with id "${id}"`);
+      return null;
+    }
```
bvaughn: Can we combine these two warnings?

```js
let fiber = idToFiberMap.get(id);

if (fiber != null) {
  // Find the currently mounted version of this fiber (so we don't show the wrong props and state).
  fiber = findCurrentFiberUsingSlowPath(fiber);
}

if (fiber == null) {
  console.warn(`Could not find Fiber with id "${id}"`);
  return null;
}
```

src/backend/renderer.js
```diff
@@ -1729,13 +1764,21 @@ export function attach(
     }
   }
 
+  function findCurrentFiberUsingSlowPathById(id: number): Fiber | null {
+    const fiber = idToFiberMap.get(id);
+    if (fiber) {
+      return findCurrentFiberUsingSlowPath(fiber);
+    }
+    return null;
+  }
```
bvaughn: Looks like there aren't really any uses of `findCurrentFiberUsingSlowPath()` anymore. Let's just change that method to accept an `id` instead and we can get rid of this indirection here (and in `inspectElementRaw()`)

sompylasar: I'd be down if we also rename the original `findCurrentFiberUsingSlowPath` to `...ById` then to highlight that it's different from what is copied from React.

.flowconfig
```diff
@@ -26,6 +26,7 @@ shells/dev/build/*
 
 [options]
 esproposal.class_instance_fields=enable
+esproposal.optional_chaining=enable
```
sompylasar: I wonder why is this still here, I remember reverting because it broke compilation of tests.

src/backend/renderer.js
bvaughn: Yeah, the name change would seem appropriate to match our existing conventions too.

bvaughn:
I'm going to resolve conflicts and land this.

sompylasar:
Thanks!

PR #241, opened 2019-04-28
Fix ProfilingSummary type mismatch Frontend vs Backend

sompylasar:
The error was silenced via typecast through `any`.

The mismatch was: Map vs Array in `initialTreeBaseDurations`.

### DIFF
--- a/src/devtools/ProfilingCache.js
+++ b/src/devtools/ProfilingCache.js
@@ -208,13 +208,10 @@ export default class ProfilingCache {
       return new Promise(resolve => {
         const importedProfilingData = this._store.importedProfilingData;
         if (importedProfilingData !== null) {
-          const { profilingSummary } = (importedProfilingData: any);
-          if (profilingSummary != null) {
-            this._pendingProfileSummaryMap.set(
-              profilingSummary.rootID,
-              resolve
-            );
-            this.onProfileSummary(profilingSummary);
+          const profilingSummaryFrontend =
+            importedProfilingData.profilingSummary;
+          if (profilingSummaryFrontend != null) {
+            resolve(profilingSummaryFrontend);
             return;
           }
         } else if (this._store.profilingOperations.has(rootID)) {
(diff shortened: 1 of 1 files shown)


sompylasar:
Related, maybe merge this one there: https://github.com/bvaughn/react-devtools-experimental/pull/248

src/devtools/ProfilingCache.js
```diff
@@ -365,32 +362,42 @@ export default class ProfilingCache {
...
-        interactionCount,
-      });
+      resolve(
+        convertProfilingSummaryFromBackendToFrontend(profilingSummaryBackend)
+      );
     }
   };
 }
+
+function convertProfilingSummaryFromBackendToFrontend(
```
bvaughn: Why split this out? It's only used/called in one place.

src/devtools/ProfilingCache.js
```diff
@@ -208,13 +208,10 @@ export default class ProfilingCache {
...
-          if (profilingSummary != null) {
-            this._pendingProfileSummaryMap.set(
-              profilingSummary.rootID,
-              resolve
-            );
-            this.onProfileSummary(profilingSummary);
+          const profilingSummaryFrontend =
+            importedProfilingData.profilingSummary;
+          if (profilingSummaryFrontend != null) {
+            resolve(profilingSummaryFrontend);
```
bvaughn: Nice fix.

sompylasar:
Yes, it would make sense to merge.

src/devtools/ProfilingCache.js
sompylasar: > Why split this out? It's only used/called in one place.

Make the intent clear. This is a separate (testable btw) piece of logic. We're splitting into functions to make the code more readable and manageable, not just for DRY.

bvaughn: I don't think splitting things into helper functions _necessarily_ improves readability or maintainability. (In this case, I don't think it does.) And it seems sufficient to test this with a more end-to-end test (as part of testing the whole import functionality) rather than in a one-off test.

PR #242, opened 2019-04-28
Improve Flow coverage by reducing typecasts through `any`

sompylasar:
There are many unnecessary typecasts through `any` which
break the Flow of types across the program.

It's more bulletproof to avoid lying to ourselves about types.

Also fixed sketchy null check where zero ID would be skipped:
```diff
-  } else if (selectedElementID) {
+  } else if (selectedElementID !== null) {
```


src/devtools/store.js
```diff
@@ -498,20 +498,26 @@ export default class Store extends EventEmitter {
     return list;
   }
 
-  getRendererIDForElement(id: number): number | null {
+  getRendererIDForElement(id: number | null): number | null {
```
bvaughn: I don't think I agree with this type change. It makes no sense to call `getRendererIDForElement` with a null value. There could be no renderer ID for a non-element.

src/devtools/store.js
```diff
@@ -498,20 +498,26 @@ export default class Store extends EventEmitter {
...
+        return rendererID == null ? null : rendererID;
       } else {
         current = this._idToElement.get(current.parentID);
       }
     }
     return null;
   }
 
-  getRootIDForElement(id: number): number | null {
+  getRootIDForElement(id: number | null): number | null {
```
bvaughn: Same as the above comment.

src/devtools/views/Profiler/ChartNode.js
```diff
@@ -10,8 +10,8 @@ type Props = {|
   height: number,
   isDimmed?: boolean,
   label: string,
-  onClick: Function,
-  onDoubleClick?: Function,
+  onClick: (event: SyntheticMouseEvent<*>) => mixed,
+  onDoubleClick?: (event: SyntheticMouseEvent<*>) => mixed,
```
bvaughn: This change looks 👍

src/devtools/store.js
```diff
@@ -523,7 +529,10 @@ export default class Store extends EventEmitter {
     return null;
   }
 
-  isInsideCollapsedSubTree(id: number): boolean {
+  isInsideCollapsedSubTree(id: number | null): boolean {
```
bvaughn: Same.

src/devtools/views/Profiler/FilterModal.js
```diff
@@ -19,8 +19,8 @@ export default function FilterModal({ dismissModal }: Props) {
   } = useContext(ProfilerContext);
 
   const handleNumberChange = useCallback(
-    ({ currentTarget }) => {
-      const newValue = parseFloat(currentTarget.value);
+    (event: SyntheticEvent<HTMLInputElement>) => {
```
bvaughn: nit: `({ currentTarget }: SyntheticEvent<HTMLInputElement>)` ?

src/devtools/views/Profiler/FilterModal.js
```diff
@@ -29,9 +29,10 @@ export default function FilterModal({ dismissModal }: Props) {
   );
 
   const handleEnabledChange = useCallback(
-    ({ currentTarget }) => {
-      setIsCommitFilterEnabled(currentTarget.checked);
-      if (currentTarget.checked) {
+    (event: SyntheticEvent<HTMLInputElement>) => {
```
bvaughn: Same

src/devtools/views/Profiler/ProfilerContext.js
```diff
@@ -115,18 +115,15 @@ function ProfilerContextController({ children }: Props) {
...
   } else if (selectedElementID) {
-    rendererID = store.getRendererIDForElement(
-      ((selectedElementID: any): number)
-    );
-    rootID = store.getRootIDForElement(((selectedElementID: any): number));
-    rootHasProfilingData = store.profilingOperations.has(
-      ((rootID: any): number)
-    );
+    rendererID = store.getRendererIDForElement(selectedElementID);
+    rootID = store.getRootIDForElement(selectedElementID);
```
bvaughn: Seems like the null checks should be done here rather than in the above signatures.

src/devtools/views/Profiler/FilterModal.js
sompylasar: It's changed for uniformity and familiarity, `event.currentTarget` is a noticeable pattern. Destructuring in this case is not.

src/devtools/views/Profiler/ProfilerContext.js
sompylasar: OK.

sompylasar:
Updated, resolved conflicts. Did not revert the `{ currentTarget }` to `event.currentTarget` change for uniformity and familiarity, I believe `event.currentTarget` is a noticeable pattern whereas destructuring in this case is not.

bvaughn:
> Updated, resolved conflicts. Did not revert the `{ currentTarget }` to `event.currentTarget` change for uniformity and familiarity, I believe `event.currentTarget` is a noticeable pattern whereas destructuring in this case is not.

This is subjective 😄 I think there's something to be said to following established conventions within a codebase, and the destructuring of `({ currentTarget })` occurs ~20 places in this project.

That being said, I'm not willing to die on this hill. I don't promise not to change it back though.

sompylasar:
> I think there's something to be said to following established conventions within a codebase, and the destructuring of ({ currentTarget }) occurs ~20 places in this project.

Yes, thanks for pointing out.

Well, as my father used to say (~roughly, English translation doesn't sound so funny) "My business is to suggest, your business is to refuse".

bvaughn:
😆

PR #243, opened 2019-04-28
Rename Profiler types to disambiguate which is Frontend and Backend

sompylasar:
Flow IDE shows the original type, not the file-scoped alias,
so it's confusing to see incompatible variables that have types
with the same name.

This started from [`src/devtools/ProfilingCache.js` where
`ProfilingSummary` was incompatible](https://github.com/bvaughn/react-devtools-experimental/pull/241), but I decided to make a bigger
change for consistency and to avoid similar mistakes in the future.

### DIFF
--- a/src/devtools/views/Profiler/utils.js
+++ b/src/devtools/views/Profiler/utils.js
@@ -1,6 +1,6 @@
 // @flow
 
-import type { CommitDetails, CommitTree, Node } from './types';
+import type { CommitDetailsFrontend, CommitTreeFrontend } from './types';
 
 const commitGradient = [
   'var(--color-commit-gradient-0)',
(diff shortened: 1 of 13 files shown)


sompylasar:
Related, maybe merge this one there: https://github.com/bvaughn/react-devtools-experimental/pull/248

src/devtools/views/Profiler/utils.js
```diff
@@ -28,8 +28,11 @@ export const calculateSelfDuration = (
   }
 
   let selfDuration = ((actualDurations.get(id): any): number);
+  const node = nodes.get(id);
+  if (node == null) {
+    throw Error(`Could not find node with id "${id}" in commit tree`);
+  }
```
bvaughn: FWIW, I didn't add explicit checks for things like this initially because the line immediately below (`node.children`) would error anyway.

sompylasar: > FWIW, I didn't add explicit checks for things like this initially because the line immediately below (node.children) would error anyway.

I prefer to not leave a possibility for obscure errors like "cannot read property children of null" if I can. And with Flow/TypeScript I can.

PR #245, opened 2019-04-29
Add configurable component filters

### LINKED ISSUE #231: Add configurable filter preferences
### Goals
* Filtering preferences need to support both temporary values (e.g. things you do while debugging- like showing only `Suspense` nodes) and persisted values (e.g. things you have a subjective standing preference for- like always showing host nodes or always hiding context providers).
* Preferences will require a format that can be shared between the frontend and backend. (The backend will need to be able to read the preference values during startup, so that it can properly filter the tree, before the frontend has even been rendered.)
* Filtering should be powerful enough to also eventually support user-specified, custom filters (e.g. hide all components that start with a certain name, or within a specific folder on the file system).
* Filtering changes should be able to be applied without requiring a reload of the app. (Toggling the `Suspense`-only mode should be doable without losing your app's current state.) For this to work, we'll need to clear all of the `Store`'s current nodes (and notify contexts like `TreeContext`), then request the backend to re-crawl and initialize the tree with the new filter criteria in mind.

### UI/UX
I think we'll probably want to add some sort of quick toggle to the Components tree for "suspense only" mode, but regarding the interface for "persisted preferences", I'm imagining something vaguely like this:

```
(✓) Hide host nodes (e.g. <div>)
( ) Hide context providers
---
[-] Hide components in "node_modules/react-relay"
[-] Hide components with names that begin with "Redux"
[+] Add new filter
```
Where the top two options are examples of hard-coded preferences we might add, and the bottom rows are examples of user-defined filters we might support. (Not sure how many options we'll provide for this initially.)

We _could_ potentially make the host nodes and context provides a user-defined filter of "hide components by type" too. Not sure. I think we'll want to factor in how we handle default settings for these preferences.

The temporary "suspense only" preference would essentially override all of the above preferences and show only Suspense nodes. (Even if the user had added a conflicting custom filter to e.g. hide all suspense nodes).

### Open questions
* Should filtering affect the Profiler and its graphs?

### Other
Related earlier proof of concept: https://github.com/facebook/react-devtools/pull/1240

bvaughn:
Resolves #231

### Feature checklist
- [x] Add support for configurable element type, display name, and location filters.
- [x] Add UI for editing filters.
- [x] Add unit tests for filters.

### Bug checklist
- [x] Profiler behaves strangely if you filter out certain nodes (e.g. `^List$` in the dev shell). In this case, the flame graph is sometimes unexpectedly truncated after the root node.
- [x] ~~Updating filters sometimes logs hundreds of "_Could not find Fiber with id_" warnings inside of the `selectElement` function during which time the extension UI becomes unresponsive.~~ Edit: I am no longer able to reproduce this error. Maybe I fixed it? We should keep an eye out for it in case.

### Demo (Components tree)
more-filter-demo-Kapture 2019-05-01 at 14 26 18

### Demo (Profiler)
aaa-profil-Kapture 2019-05-01 at 17 20 15


bvaughn:
Progress update...
filter-demo-Kapture 2019-05-01 at 9 26 16

bvaughn:
This code can always be cleaned up more, and  I'm sure a few things will break, but I'd like to get people playing with this feature. I'll be ready to be in bug-fixing mode as reports come in.

sompylasar:
This looks awesome, thanks!

bvaughn:
Thank you, @sompylasar

Issue #232, opened 2019-04-26
Clear profiling data for root when the root is unmounted


bvaughn:
See PR #230

We currently only reset profiling data when profiling is re-started (in `onProfilingStatus`)
https://github.com/bvaughn/react-devtools-experimental/blob/c28acc5d471883a19ca562365f4974c36da48dd3/src/devtools/store.js#L965-L972

We should probably also clear it out if the root it corresponds to gets unmounted.

This should be easy for `_profilingOperations` and `_profilingSnapshot` since they are just maps of root ID to data.

However, I think `_profilingScreenshots` is actually currently broken because it maps commit index to screenshot, rather than root id to commit index to screenshot. So we should fix that too!

A basic test or two that verifies this would be nicer. 😄

PR #248, opened 2019-04-30
Cleanup profile export/import data types, add basic tests

sompylasar:
Cleanup profile export/import data types, add export/import test

Strengthened the I/O-boundary type conversion logic.

Fixed type inconsistencies uncovered by removing `any` and
by making explicit type annotations and transformations.

In particular, these were likely malformed when restored from a file:
- `commitDetails`
- `interactions`
- `initialTreeBaseDurations`

The mismatches were Map vs interleaved Array.

### DIFF
--- a/src/__tests__/__snapshots__/profiling-test.js.snap
+++ b/src/__tests__/__snapshots__/profiling-test.js.snap
@@ -78,81 +78,74 @@ Object {
 }
 `;
 
-exports[`profiling CommitDetails should be collected for each commit: exported data 1`] = `
+exports[`profiling CommitDetails should be collected for each commit: imported data 1`] = `
 Object {
   "commitDetails": Array [
     Object {
+      "actualDurations": Map {
+        1 => 12,
+        2 => 12,
+        3 => 0,
+        4 => 1,
+        5 => 1,
+      },
       "commitIndex": 0,
-      "durations": Array [
-        1,
-        12,
-        0,
-        2,
-        12,
-        10,
-        3,
-        0,
-        0,
-        4,
-        1,
-        1,
-        5,
-        1,
-        1,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        1 => 0,
+        2 => 10,
+        3 => 0,
+        4 => 1,
+        5 => 1,
+      },
     },
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        4 => 1,
+        6 => 2,
+        2 => 13,
+        1 => 13,
+      },
       "commitIndex": 1,
-      "durations": Array [
-        3,
-        0,
-        0,
-        4,
-        1,
-        1,
-        6,
-        2,
-        2,
-        2,
-        13,
-        10,
-        1,
-        13,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        4 => 1,
+        6 => 2,
+        2 => 10,
+        1 => 0,
+      },
     },
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        2 => 10,
+        1 => 10,
+      },
       "commitIndex": 2,
-      "durations": Array [
-        3,
-        0,
-        0,
-        2,
-        10,
-        10,
-        1,
-        10,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        2 => 10,
+        1 => 0,
+      },
     },
     Object {
+      "actualDurations": Map {
+        2 => 10,
+        1 => 10,
+      },
       "commitIndex": 3,
-      "durations": Array [
-        2,
-        10,
-        10,
-        1,
-        10,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        2 => 10,
+        1 => 0,
+      },
     },
   ],
   "interactions": Object {
@@ -161,7 +154,7 @@ Object {
   },
   "profilingOperations": Map {
     1 => Array [
-      Array [
+      Uint32Array [
         1,
         1,
         17,
@@ -231,7 +224,7 @@ Object {
         5,
         1000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         8,
@@ -267,7 +260,7 @@ Object {
         1,
         14000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         0,
@@ -287,7 +280,7 @@ Object {
         1,
         11000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         0,
@@ -313,7 +306,7 @@ Object {
       35,
       45,
     ],
-    "initialTreeBaseDurations": Array [],
+    "initialTreeBaseDurations": Map {},
     "interactionCount": 0,
     "rootID": 1,
   },
@@ -442,68 +435,61 @@ Object {
 }
 `;
 
-exports[`profiling FiberCommits should be collected for each rendered fiber: exported data 1`] = `
+exports[`profiling FiberCommits should be collected for each rendered fiber: imported data 1`] = `
 Object {
   "commitDetails": Array [
     Object {
+      "actualDurations": Map {
+        1 => 11,
+        2 => 11,
+        3 => 0,
+        4 => 1,
+      },
       "commitIndex": 0,
-      "durations": Array [
-        1,
-        11,
-        0,
-        2,
-        11,
-        10,
-        3,
-        0,
-        0,
-        4,
-        1,
-        1,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        1 => 0,
+        2 => 10,
+        3 => 0,
+        4 => 1,
+      },
     },
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        5 => 1,
+        2 => 11,
+        1 => 11,
+      },
       "commitIndex": 1,
-      "durations": Array [
-        3,
-        0,
-        0,
-        5,
-        1,
-        1,
-        2,
-        11,
-        10,
-        1,
-        11,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        5 => 1,
+        2 => 10,
+        1 => 0,
+      },
     },
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        5 => 1,
+        6 => 2,
+        2 => 13,
+        1 => 13,
+      },
       "commitIndex": 2,
-      "durations": Array [
-        3,
-        0,
-        0,
-        5,
-        1,
-        1,
-        6,
-        2,
-        2,
-        2,
-        13,
-        10,
-        1,
-        13,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        5 => 1,
+        6 => 2,
+        2 => 10,
+        1 => 0,
+      },
     },
   ],
   "interactions": Object {
@@ -512,7 +498,7 @@ Object {
   },
   "profilingOperations": Map {
     1 => Array [
-      Array [
+      Uint32Array [
         1,
         1,
         15,
@@ -570,7 +556,7 @@ Object {
         4,
         1000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         8,
@@ -605,7 +591,7 @@ Object {
         1,
         12000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         8,
@@ -657,7 +643,7 @@ Object {
       22,
       35,
     ],
-    "initialTreeBaseDurations": Array [],
+    "initialTreeBaseDurations": Map {},
     "interactionCount": 0,
     "rootID": 1,
   },
@@ -666,47 +652,42 @@ Object {
 `;
 
 exports[`profiling Interactions should be collected for every traced interaction: Interactions 1`] = `
-Array [
-  Object {
-    "__count": 1,
-    "commits": Array [
-      0,
-    ],
-    "id": 0,
-    "name": "mount: one child",
-    "timestamp": 0,
-  },
-  Object {
-    "__count": 0,
-    "commits": Array [
-      1,
-    ],
-    "id": 1,
-    "name": "update: two children",
-    "timestamp": 11,
-  },
-]
+Object {
+  "interactions": Array [
+    Object {
+      "__count": 1,
+      "commits": Array [
+        0,
+      ],
+      "id": 0,
+      "name": "mount: one child",
+      "timestamp": 0,
+    },
+    Object {
+      "__count": 0,
+      "commits": Array [
+        1,
+      ],
+      "id": 1,
+      "name": "update: two children",
+      "timestamp": 11,
+    },
+  ],
+  "rootID": 1,
+}
 `;
 
-exports[`profiling Interactions should be collected for every traced interaction: exported data 1`] = `
+exports[`profiling Interactions should be collected for every traced interaction: imported data 1`] = `
 Object {
   "commitDetails": Array [
     Object {
+      "actualDurations": Map {
+        1 => 11,
+        2 => 11,
+        3 => 0,
+        4 => 1,
+      },
       "commitIndex": 0,
-      "durations": Array [
-        1,
-        11,
-        0,
-        2,
-        11,
-        10,
-        3,
-        0,
-        0,
-        4,
-        1,
-        1,
-      ],
       "interactions": Array [
         Object {
           "__count": 1,
@@ -716,23 +697,21 @@ Object {
         },
       ],
       "rootID": 1,
+      "selfDurations": Map {
+        1 => 0,
+        2 => 10,
+        3 => 0,
+        4 => 1,
+      },
     },
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        5 => 1,
+        2 => 11,
+        1 => 11,
+      },
       "commitIndex": 1,
-      "durations": Array [
-        3,
-        0,
-        0,
-        5,
-        1,
-        1,
-        2,
-        11,
-        10,
-        1,
-        11,
-        0,
-      ],
       "interactions": Array [
         Object {
           "__count": 0,
@@ -742,6 +721,12 @@ Object {
         },
       ],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        5 => 1,
+        2 => 10,
+        1 => 0,
+      },
     },
   ],
   "interactions": Object {
@@ -769,7 +754,7 @@ Object {
   },
   "profilingOperations": Map {
     1 => Array [
-      Array [
+      Uint32Array [
         1,
         1,
         15,
@@ -827,7 +812,7 @@ Object {
         4,
         1000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         8,
@@ -876,7 +861,7 @@ Object {
       11,
       22,
     ],
-    "initialTreeBaseDurations": Array [],
+    "initialTreeBaseDurations": Map {},
     "interactionCount": 2,
     "rootID": 1,
   },
@@ -908,59 +893,55 @@ Object {
 }
 `;
 
-exports[`profiling ProfilingSummary should be collected for each commit: exported data 1`] = `
+exports[`profiling ProfilingSummary should be collected for each commit: imported data 1`] = `
 Object {
   "commitDetails": Array [
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        4 => 1,
+        6 => 2,
+        2 => 13,
+        1 => 13,
+      },
       "commitIndex": 0,
-      "durations": Array [
-        3,
-        0,
-        0,
-        4,
-        1,
-        1,
-        6,
-        2,
-        2,
-        2,
-        13,
-        10,
-        1,
-        13,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        4 => 1,
+        6 => 2,
+        2 => 10,
+        1 => 0,
+      },
     },
     Object {
+      "actualDurations": Map {
+        3 => 0,
+        2 => 10,
+        1 => 10,
+      },
       "commitIndex": 1,
-      "durations": Array [
-        3,
-        0,
-        0,
-        2,
-        10,
-        10,
-        1,
-        10,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        3 => 0,
+        2 => 10,
+        1 => 0,
+      },
     },
     Object {
+      "actualDurations": Map {
+        2 => 10,
+        1 => 10,
+      },
       "commitIndex": 2,
-      "durations": Array [
-        2,
-        10,
-        10,
-        1,
-        10,
-        0,
-      ],
       "interactions": Array [],
       "rootID": 1,
+      "selfDurations": Map {
+        2 => 10,
+        1 => 0,
+      },
     },
   ],
   "interactions": Object {
@@ -969,7 +950,7 @@ Object {
   },
   "profilingOperations": Map {
     1 => Array [
-      Array [
+      Uint32Array [
         1,
         1,
         8,
@@ -1005,7 +986,7 @@ Object {
         1,
         14000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         0,
@@ -1025,7 +1006,7 @@ Object {
         1,
         11000,
       ],
-      Array [
+      Uint32Array [
         1,
         1,
         0,
@@ -1091,18 +1072,13 @@ Object {
       23,
       33,
     ],
-    "initialTreeBaseDurations": Array [
-      1,
-      12,
-      2,
-      12,
-      3,
-      0,
-      4,
-      1,
-      5,
-      1,
-    ],
+    "initialTreeBaseDurations": Map {
+      1 => 12,
+      2 => 12,
+      3 => 0,
+      4 => 1,
+      5 => 1,
+    },
     "interactionCount": 0,
     "rootID": 1,
   },
--- a/src/__tests__/profiling-test.js
+++ b/src/__tests__/profiling-test.js
@@ -14,46 +14,6 @@ describe('profiling', () => {
   let store: Store;
   let utils;
 
-  const exportImportHelper = (rendererID: number, rootID: number) => {
-    const {
-      prepareProfilingExport,
-      prepareProfilingImport,
-    } = require('src/devtools/views/Profiler/utils');
-
-    let exportedProfilingSummary;
-    bridge.addListener('exportFile', ({ contents }) => {
-      exportedProfilingSummary = contents;
-    });
-
-    utils.act(() => {
-      const exportProfilingSummary = prepareProfilingExport(
-        store.profilingOperations,
-        store.profilingSnapshots,
-        rootID,
-        rendererID
-      );
-      bridge.send('exportProfilingSummary', exportProfilingSummary);
-    });
-
-    expect(exportedProfilingSummary).toBeDefined();
-
-    const importedProfilingSummary = prepareProfilingImport(
-      ((exportedProfilingSummary: any): string)
-    );
-
-    // Sanity check that profiling snapshots are serialized correctly.
-    expect(store.profilingSnapshots.get(rootID)).toEqual(
-      importedProfilingSummary.profilingSnapshots.get(rootID)
-    );
-
-    // Snapshot the JSON-parsed object, rather than the raw string, because Jest formats the diff nicer.
-    expect(importedProfilingSummary).toMatchSnapshot('exported data');
-
-    utils.act(() => {
-      store.importedProfilingData = importedProfilingSummary;
-    });
-  };
-
   beforeEach(() => {
     utils = require('./utils');
     utils.beforeEachProfiling();
@@ -71,10 +31,10 @@ describe('profiling', () => {
 
   it('should throw if importing older/unsupported data', () => {
     const {
-      prepareProfilingImport,
+      prepareImportedProfilingData,
     } = require('src/devtools/views/Profiler/utils');
     expect(() =>
-      prepareProfilingImport(
+      prepareImportedProfilingData(
         JSON.stringify({
           version: 0,
         })
@@ -143,7 +103,7 @@ describe('profiling', () => {
 
       expect(profilingSummary).not.toBeNull();
 
-      exportImportHelper(rendererID, rootID);
+      utils.exportImportHelper(bridge, store, rendererID, rootID);
 
       await utils.actAsync(() =>
         TestRenderer.create(
@@ -234,7 +194,7 @@ describe('profiling', () => {
 
       expect(allCommitDetails).toHaveLength(4);
 
-      exportImportHelper(rendererID, rootID);
+      utils.exportImportHelper(bridge, store, rendererID, rootID);
 
       for (let commitIndex = 0; commitIndex < 4; commitIndex++) {
         await utils.actAsync(() => {
@@ -466,7 +426,7 @@ describe('profiling', () => {
 
       expect(allFiberCommits).toHaveLength(store.numElements);
 
-      exportImportHelper(rendererID, rootID);
+      utils.exportImportHelper(bridge, store, rendererID, rootID);
 
       for (let index = 0; index < store.numElements; index++) {
         await utils.actAsync(() => {
@@ -562,7 +522,7 @@ describe('profiling', () => {
 
       expect(interactions).not.toBeNull();
 
-      exportImportHelper(rendererID, rootID);
+      utils.exportImportHelper(bridge, store, rendererID, rootID);
 
       await utils.actAsync(() =>
         TestRenderer.create(
--- a/src/__tests__/setupTests.js
+++ b/src/__tests__/setupTests.js
@@ -31,31 +31,43 @@ env.beforeEach(() => {
 
   installHook(global);
 
-  const bridgeListeners = [];
-  const bridge = new Bridge({
-    listen(callback) {
-      bridgeListeners.push(callback);
-      return () => {
-        const index = bridgeListeners.indexOf(callback);
-        if (index >= 0) {
-          bridgeListeners.splice(index, 1);
-        }
-      };
-    },
-    send(event: string, payload: any, transferable?: Array<any>) {
-      bridgeListeners.forEach(callback => callback({ event, payload }));
-    },
-  });
+  function init() {
+    const bridgeListeners = [];
+    const bridge = new Bridge({
+      listen(callback) {
+        bridgeListeners.push(callback);
+        return () => {
+          const index = bridgeListeners.indexOf(callback);
+          if (index >= 0) {
+            bridgeListeners.splice(index, 1);
+          }
+        };
+      },
+      send(event: string, payload: any, transferable?: Array<any>) {
+        bridgeListeners.forEach(callback => callback({ event, payload }));
+      },
+    });
 
-  const agent = new Agent(bridge);
+    const agent = new Agent(bridge);
 
-  const hook = global.__REACT_DEVTOOLS_GLOBAL_HOOK__;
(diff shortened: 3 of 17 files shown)


sompylasar:
P.S. I'd love to not include huge JSONs and snapshots, but I'm not sure how to get a smaller profiling session recording.

sompylasar:
Related, maybe merge here: https://github.com/bvaughn/react-devtools-experimental/pull/241

bvaughn:
> In particular, these were likely malformed when restored from a file:
* commitDetails
* interactions

Could you elaborate? I was able to export and import profiling data. Seems like I wouldn't have been able to if the data was "malformed".

bvaughn:
> Found out that there's no single serializable "profiling snapshot" type
that is created in one place and can be saved into a file,
instead it is assembled from pieces coming from different sources.
I'd appreciate some feedback about the intent behind these pieces
and then I'd like this cleaned up even more, if possible.

Have you read the [Profiling overview doc](https://github.com/bvaughn/react-devtools-experimental/blob/master/OVERVIEW.md#profiler)? The intent behind the current data structure is to enable it to be lazily sent across the bridge as needed by the different Profiler UI states.

For that matter, there _is_ a single type. It's what's saved to JSON when you export. It's what you named (in this PR) `ExportedProfilingDataFromRendererV1`.

> P.S. I'd love to not include huge JSONs and snapshots, but I'm not sure how to get a smaller profiling session recording.

What's "huge" about this data? Be specific. I think it's pretty streamlined to enable us to lazily compute the bits that we need to make the frontend/UI work.

src/backend/types.js
```diff
@@ -90,6 +90,17 @@ export type ProfilingSummary = {|
   rootID: number,
 |};
 
+type ExportedProfilingDataFromRendererV1 = {|
```
bvaughn: I think `ExportedProfilingData` as a type with a `version: number` field makes more sense.

I don't _think_ we'll necessarily want to support multiple versions. If we add new fields or change the data in some meaningful way, that will break old exports.

src/devtools/views/Profiler/__tests__/utils-test.js
```diff
@@ -0,0 +1,53 @@
...
+        })
+      ).toMatchSnapshot();
+    });
+  });
+
+  describe('parseExportedProfilingDataString', () => {
+    it('works', () => {
+      expect(
+        // This is a recording of `shells/dev/index.html`
+        parseExportedProfilingDataString(`
```
bvaughn: Wouldn't it be sufficient (and easier to read) to just define some profiling data, export, re-import, and compare to original? What's the function of this hard-coded blob of JSON?

sompylasar: > > P.S. I'd love to not include huge JSONs and snapshots, but I'm not sure how to get a smaller profiling session recording.
> 
> What's "huge" about this data? Be specific. I think it's pretty streamlined to enable us to lazily compute the bits that we need to make the frontend/UI work.

Sorry for not making it very clear, I meant this JSON in the test.

> Wouldn't it be sufficient (and easier to read) to just define some profiling data, export, re-import, and compare to original? What's the function of this hard-coded blob of JSON?

Sure (although I wanted this test to not depend on the whole Store/Bridge infra), by "just define some profiling data" what do you mean specifically?

src/devtools/views/Profiler/utils.js
```diff
@@ -68,3 +75,128 @@ export const scale = (
...
+  };
+  return importedProfileData;
+}
+
+export function parseExportedProfilingDataString(
+  str: string
+): ImportedProfilingData {
+  const jsonData = JSON.parse(str);
+  // When adding newer versions, add a parser into this map.
+  const parsers = new Map([[1, parseExportedProfilingDataV1]]);
```
bvaughn: This is an interesting idea. I'm not sure if it's worth doing though. Seems like too many changes wouldn't be backwards compat without complicating the UI.

sompylasar:
> > In particular, these were likely malformed when restored from a file:
> > 
> > - commitDetails
> > - interactions
>
> Could you elaborate? I was able to export and import profiling data. Seems like I wouldn't have been able to if the data was "malformed".

I don't know and haven't tested how it worked before the change (just had no extra time) but from my intuition I believe it was deserialized into wrong data types which may have worked by accident in runtime. I can find out what actually gets there and how could it work without matching the types.

src/devtools/views/Profiler/__tests__/utils-test.js
bvaughn: > Sorry for not making it very clear, I meant this JSON in the test.

Oh! That makes a lot more sense now. Thanks for clarifying 😄 

> Sure (although I wanted this test to not depend on the whole Store/Bridge infra), by "just define some profiling data" what do you mean specifically?

Yeah, sorry. My original comment was not very clear. I was just thinking we could combine the two above tests into a single one that only has one hard-coded source of data.

1. Start with a hard-coded object of profiler data (this part is still yuck)
1. Serialize it using `prepareExportedProfilingSummaryFromFrontend` then `JSON.stringify`
1. Reconstruct it using `parseExportedProfilingDataV1`
1. Expect to be shallow-equal to the original

Would probably be better still if we could generate *real* profiling data though, as hard-coded mocks can get stale easily.

src/backend/types.js
sompylasar: My idea here is to express the intent with data type names very clearly. Even if you won't support multiple versions in the same codebase, you may still want to know what is type-compatible with what, so that the compiler tells you what you might have broken. Aa you might have noticed, this type is not exported (on purpose; I should've left a comment about my design decision here).

bvaughn: Not sure I see the value of that type, since there's no way to enforce it. (We would still need a runtime check of the data version to make sure it's the one the current build supports. This data comes from an external source; it isn't defined in a Flow parsed file)

src/devtools/views/Profiler/__tests__/utils-test.js
sompylasar: Oops accidentally resolved, github made "reply" not a button. Unresolved now.

sompylasar: As I remember from last night that these two things `prepareExportedProfilingSummaryFromFrontend` and `parseExportedProfilingDataV1` aren't type-compatible. But I need to take a fresh look.

src/devtools/views/Profiler/utils.js
sompylasar: It's okay to keep just one here if you want to. But make the intent very clear using names, not assumed.

src/devtools/views/Profiler/__tests__/utils-test.js
bvaughn: Conceptually, all I'm saying is that it would be nice if we had a single test that verified that we could start with some known profiling data<sup>1</sup>, serialize it, then deserialize it and end up with the same data.

<sup>1</sup> Preferably *real* data rather than hard-coded data.

src/devtools/views/Profiler/utils.js
bvaughn: I never advocated for "assuming" 😄 We need a runtime version check (as the TODO that was in place before called for)

>  // TODO (profiling) Version check; warn if older version.

src/backend/types.js
sompylasar: > Not sure I see the value of that type, since there's no way to enforce it.

The value is that you describe what is expected there. I'm coming from TypeScript background where I had `io-ts` for runtime type enforcement and derived compile-time types. I think it has Flow, too, if you want to look into making I/O boundaries typed.

> (We would still need a runtime check of the data version to make sure it's the one the current build supports. This data comes from an external source; it isn't defined in a Flow parsed file)

You're right. In my vision there's a clear boundary (I call it I/O boundary) between the untyped world (JSON, files) and typed world (Flow), and there should definitely be functions that ensure the types and pipe data from untyped world to typed, serve as a single entry point into the typed world.

src/devtools/views/Profiler/__tests__/utils-test.js
sompylasar: Yes, I agree. But this would require spinning up the whole infra, rendering the app, profiling it, exporting, importing. I agree such a test would be more valuable than this one.

src/devtools/views/Profiler/utils.js
sompylasar: I removed the TODO because I built the version check and some of the deep structure checks.

sompylasar:
Off-topic, but speaking of benefits of Flow types and not lying to yourself with `any`, I'm writing a proper test for the export/import and just hit this:
```
  console.warn src/backend/agent.js:170
    Unable to export file: TypeError: Cannot read property 'get' of null
        at get (/Users/sompylasar/_github/react-devtools-experimental/src/backend/renderer.js:1944:107)
        at Object.getProfilingSummary [as getExportedProfilingData] (/Users/sompylasar/_github/react-devtools-experimental/src/backend/renderer.js:1933:25)
        at Bridge.getExportedProfilingData (/Users/sompylasar/_github/react-devtools-experimental/src/backend/agent.js:158:58)
        at Bridge.emit (events.js:182:13)
        at emit (/Users/sompylasar/_github/react-devtools-experimental/src/bridge.js:28:14)
        at callback (/Users/sompylasar/_github/react-devtools-experimental/src/__tests__/setupTests.js:43:45)
        at Array.forEach (<anonymous>)
        at Object.forEach [as send] (/Users/sompylasar/_github/react-devtools-experimental/src/__tests__/setupTests.js:43:25)
        at send (/Users/sompylasar/_github/react-devtools-experimental/src/bridge.js:101:20)
        at _timers.set.callback (/Users/sompylasar/_github/react-devtools-experimental/node_modules/@jest/fake-timers/build/jestFakeTimers.js:505:25)
        at callback (/Users/sompylasar/_github/react-devtools-experimental/node_modules/@jest/fake-timers/build/jestFakeTimers.js:497:29)
        at FakeTimers._runTimerHandle (/Users/sompylasar/_github/react-devtools-experimental/node_modules/@jest/fake-timers/build/jestFakeTimers.js:541:9)
        at FakeTimers.runAllTimers (/Users/sompylasar/_github/react-devtools-experimental/node_modules/@jest/fake-timers/build/jestFakeTimers.js:193:12)
        at Object.runAllTimers (/Users/sompylasar/_github/react-devtools-experimental/node_modules/jest-runtime/build/index.js:1173:44)
        at runAllTimers (/Users/sompylasar/_github/react-devtools-experimental/src/__tests__/profilerExportImport-test.js:21:10)
        at Object.act (/Users/sompylasar/_github/react-devtools-experimental/src/__tests__/profilerExportImport-test.js:84:5)
        at Object.asyncJestTest (/Users/sompylasar/_github/react-devtools-experimental/node_modules/jest-jasmine2/build/jasmineAsyncInstall.js:102:37)
        at resolve (/Users/sompylasar/_github/react-devtools-experimental/node_modules/jest-jasmine2/build/queueRunner.js:43:12)
        at new Promise (<anonymous>)
        at mapper (/Users/sompylasar/_github/react-devtools-experimental/node_modules/jest-jasmine2/build/queueRunner.js:26:19)
        at promise.then (/Users/sompylasar/_github/react-devtools-experimental/node_modules/jest-jasmine2/build/queueRunner.js:73:41)
        at process._tickCallback (internal/process/next_tick.js:68:7)
```

Can you spot the problem? :)
```js
function getProfilingSummary(rootID: number): ProfilingSummaryBackend {
    const interactions = new Set();
    const commitDurations = [];
    const commitTimes = [];

    const commitProfilingMetadata = ((rootToCommitProfilingMetadataMap: any): CommitProfilingMetadataMap).get(
      rootID
    );
```

sompylasar:
Updated, but the test is supposed to break due to https://github.com/bvaughn/react-devtools-experimental/pull/230 https://github.com/bvaughn/react-devtools-experimental/issues/232

src/backend/types.js
bvaughn: Circling back. My concern is that I don't see the value in naming a type e.g. v1 if we only intend to support one type at any given time. I think the type name maybe muddies the water since (to me) it implies that we intend to support more than one version.

src/devtools/views/Profiler/__tests__/utils-test.js
bvaughn: I don't think I understand. We are essentially already doing most of this for our tests, and it's very valuable.

src/__tests__/profilerExportImport-test.js
```diff
@@ -0,0 +1,133 @@
...
+    );
+    expect(store).toMatchSnapshot('8: import for root 1 finished');
+    // TODO: Verify that the profiling data matches the data before export.
+
+    store.importedProfilingData = parseExportedProfilingDataString(
+      exportedFiles[1].contents
+    );
+    expect(store).toMatchSnapshot('8: import for root 2 finished');
+    // TODO: Verify that the profiling data matches the data before export.
+  });
```
bvaughn: I don't think the `Store` snapshots are adding any value here. We should be snapshotting the profiling data, right?

src/devtools/views/Profiler/__tests__/utils-test.js
sompylasar: > > Yes, I agree. But this would require spinning up the whole infra, rendering the app, profiling it, exporting, importing. I agree such a test would be more valuable than this one.
> 
> I don't think I understand. We are essentially already doing most of this for our tests, and it's very valuable.

Sorry for the embedded contradiction in my line of thought. I was just complaining, but agreed that the test should be changed.

src/__tests__/profilerExportImport-test.js
sompylasar: > I don't think the `Store` snapshots are adding any value here. We should be snapshotting the profiling data, right?

Yes, you're right. Copy-pasted, didn't spend enough time on the snapshots part of the test.

src/devtools/views/Profiler/__tests__/utils-test.js
bvaughn: No worries :)

src/backend/types.js
sompylasar: Renamed the types to not contain `V1`.

src/__tests__/profilerExportImport-test.js
sompylasar: Profiling data is flaky so cannot be snapshotted. :( TODO

bvaughn: Nah :) It's just that timing values vary slightly between runs. This isn't that surprising given that we're using high-res timestamps.

Don't worry about this though. I added tests over the weekend (PR #258) just need to get a new React canary built after https://github.com/facebook/react/pull/15567 is merged.

bvaughn:
PR #258 just landed.

Sorry for causing you merge conflicts here. If you're interested in rebasing this PR for the Flow type changes, we have decent profiling test coverage now. If you aren't interested, let me know and I can pick up the work

sompylasar:
I'm interested but I'm doing this off-work if I have energy. Please pick up if you have dedicated time for it. Thanks!

bvaughn:
Understood

src/devtools/views/Profiler/utils.js
```diff
@@ -72,3 +79,183 @@ export const scale = (
...
+              x =>
+                Array.isArray(x) &&
+                x.length === 2 &&
+                typeof x[0] === 'number' &&
+                Array.isArray(x[1])
+            )
+          )
+        ) {
+          throw new Error(
+            'Malformed profile data version 1: profilingSnapshotsByRootID pair'
```
bvaughn: This validation feels really heavy-handed, to the point of being hard to read.

I think I'd rather rely on our unit tests to ensure that exporting and re-importing data results in valid data, rather than having this amount of runtime validation. I think runtime validation should maybe be as simple as verifying the expected version number. (If people manually edit exported profiling data, it will break and that's okay.)

I found a bug while testing this branch (probably exists in master too) where exporting then re-importing fails, and the "_Malformed profile data version 1: profilingSnapshotsByRootID pair_" isn't all that helpful an error message anyway 😅 

This is mostly just an "FYI" type comment. I'm going to think about this a little more tomorrow, when my brain is fresh.

sompylasar: The comment itself may be tweaked to say something obscure like "Something is wrong with your profile, we couldn't parse it". I believe runtime validation at I/O boundary is useful in addition to tests, it covers different user scenario. You might not care about it, so it's your call to remove that, as well as stick `: any` in such places where there's an assumption about a type but no formal verification. Because you don't use a type-aware runtime helper library like io-ts that generates and hides this code from you, it may be intimidating to write validations by hand, I understand.

bvaughn: It's not a matter of it being intimidating. I just don't think that the validation justifies the added complexity.

sompylasar: During my refactoring I've changed (fixed) the type structure, but didn't change the version. It's a one-off case, but in such case, even though the tests will pass as I changed them, too, the users will get obscure errors from internals of our program rather than from the I/O boundary. If we had this runtime validation, we wouldn't allow bad data into the typed program. But again, it's your call to say "let it fail on property access or on rendering, we don't really care".

bvaughn: That's _not_ what I said. In the event we change the data structure, we would bump the version number as well. The end result would be a dialog explaining that the user is loading an older, unsupported version of profiling data (rather than e.g. "_Malformed profile data version 1: profilingSnapshotsByRootID pair_")

src/__tests__/profilerExportImport-test.js
bvaughn: This is sort of what I meant:
https://github.com/bvaughn/react-devtools-experimental/blob/acdc9bae674d27fb107a8ed99680148f07cfd55d/src/__tests__/profiling-test.js#L17-L55

This code takes real profiling data, exports+imports it through our bridge (admittedly this is a little manual at the moment), then tests verify that the before and after state match.

sompylasar: Looks good. 👍

src/__tests__/__snapshots__/profiling-test.js.snap
```diff
@@ -55,63 +55,48 @@ Object {
...
-        5,
-        1,
-      ],
+      "actualDurations": Map {
+        1 => 12,
+        2 => 12,
+        3 => 0,
+        4 => 1,
+        5 => 1,
+      },
```
sompylasar: Note this snapshot change, this was the bug that I was talking about, mismatch Array vs Map.

src/__tests__/__snapshots__/profiling-test.js.snap
```diff
@@ -123,7 +108,7 @@ Object {
   },
   "profilingOperations": Map {
     1 => Array [
-      Array [
+      Uint32Array [
```
sompylasar: Also this: plain Array vs typed Uint32Array

src/__tests__/__snapshots__/profiling-test.js.snap
```diff
@@ -287,7 +272,7 @@ Object {
       35,
       45,
     ],
-    "initialTreeBaseDurations": Array [],
+    "initialTreeBaseDurations": Map {},
```
sompylasar: And this: Array vs Map

src/__tests__/profiling-test.js
```diff
@@ -16,17 +16,19 @@ describe('profiling', () => {
...
     bridge.addListener('exportFile', ({ contents }) => {
-      exportedProfilingSummary = contents;
+      if (typeof contents === 'string') {
+        exportedProfilingDataJsonString = (contents: string);
+      }
     });
 
     utils.act(() => {
-      const exportProfilingSummary = prepareProfilingExport(
+      const exportProfilingSummary = prepareExportedProfilingSummary(
```
sompylasar: Note that there are two separate types of "exported": 
- `ExportedProfilingSummaryFromFrontend` that gets created by `prepareExportedProfilingSummary` 
  (⚠️ _previously this function was named as `prepareProfilingExport` but this name was confusing because it's not an exact mirror of `prepareProfilingImport`_ ⚠️), and 
- `ExportedProfilingData` that gets built by the `exportProfilingSummary` handler in the `agent` (created by `prepareExportedProfilingData`).

And one type of "imported":
- `ImportedProfilingData`.

src/backend/renderer.js
```diff
@@ -774,30 +776,30 @@ export function attach(
...
-    const id = pendingStringTable.size + 1;
-    pendingStringTable.set(str, id);
+    const stringID = pendingStringTable.size + 1;
+    pendingStringTable.set(str, stringID);
     // The string table total length needs to account
     // both for the string length, and for the array item
     // that contains the length itself. Hence + 1.
     pendingStringTableLength += str.length + 1;
-    return id;
+    return stringID;
```
sompylasar: I renamed some `id` variables in this file to say which exactly IDs (i.e. `stringID` or `fiberID`), otherwise it was confusing.

I would further suggest to use `opaque type StringID: number`, `opaque type FiberID: number`, etc. for them (a Flow equivalent of TypeScript branded types) to reliably distinguish the specific types of numbers we deal with around the codebase.

bvaughn:
> I really hope you won't be intimidated by the (Flow-)types-first approach here @bvaughn

That's a pretty condescending statement, @sompylasar 😄

src/backend/renderer.js
bvaughn: `opaque type StringID: number = number` ?

sompylasar:
> That sounds a little condescending @sompylasar 😆 but I assume you don't mean it that way.

Yeah, of course I didn't mean that, I'm sorry! My bad choice of words.

src/backend/renderer.js
sompylasar: > `opaque type StringID: number = number` ?

Yes. This will ensure that this specific number is incompatible with other specific numbers, only with non-specific numbers. The name could be more verbose, i.e. `BridgeProtocolStringID`.

sompylasar:
> But I did think the data validation methods in this PR (originally) were too heavy-handed. It didn't seem like they added enough value over the import/export flow that was in master to be worth their added weight and complexity.

I removed the original methods, but I felt I had to keep the explicit type conversion, and that's how I pinpointed the type incompatibility bug that was hiding there without it.

> I'll review the PR again at some point if you'd like me to take another look.

Yes please! I think I did everything I wanted here, opaque types for IDs is something bigger, for another PR.

bvaughn:
> I removed the original methods, but I felt I had to keep the explicit type conversion, and that's how I pinpointed the type incompatibility bug that was hiding there without it.

Sounds good to me then! I'll review it again tomorrow.

src/backend/renderer.js
bvaughn: I'm still confused by the name "StringID" for a numeric type.

bvaughn: I get the concept of using a named type alias. I just think "string id" is a confusing name for this one :grin

sompylasar: Yes, oh well English word order is so ambiguous... can be StringTableID but it's not an ID of the table, too. I'd say it's `IdOfStringInBridgeProtocolStringTable` but nobody's going to write this, right?

bvaughn: 😆 It's an ID for items _in_ the table so I think that would still be a reasonable name.

> I'd say it's `IdOfStringInBridgeProtocolStringTable` but nobody's going to write this, right?

Definitely not! lol

src/devtools/views/Profiler/utils.js
```diff
@@ -49,54 +65,145 @@ export const calculateSelfDuration = (
...
+  const profilingOperationsExported: ExportedProfilingDataOperationsByRootID =
+    parsed.profilingOperationsByRootID;
+  const profilingSnapshotsExported: ExportedProfilingDataSnapshotsByRootID =
+    parsed.profilingSnapshotsByRootID;
+  const profilingSummaryExported: ExportedProfilingDataProfilingSummary =
+    parsed.profilingSummary;
+  const commitDetailsExported: ExportedProfilingDataCommitDetails =
+    parsed.commitDetails;
+  const interactionsExported: ExportedProfilingDataInteractions =
+    parsed.interactions;
```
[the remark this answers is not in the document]
sompylasar: They are here to describe the intent and they provide typehints on how to convert them. I probably can assume with the same result that `parsed` is an `ExportedProfilingData` as a whole.

src/utils.js
```diff
@@ -81,6 +81,50 @@ function toCodePoint(string: string) {
...
+  for (let i = 0; i < interleavedArray.length; i += 2) {
+    // Assuming `TKey` and `TValue` are actually interleaved, so forcing this assumption via `any`.
+    map.set(
+      ((interleavedArray[i]: any): TKey),
+      ((interleavedArray[i + 1]: any): TValue)
+    );
+  }
+  return map;
+}
+
```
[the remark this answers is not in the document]
sompylasar: They are to describe the intent and solidify the pattern around the code.

src/__tests__/setupTests.js
```diff
@@ -31,31 +31,43 @@ env.beforeEach(() => {
...
-  global.bridge = bridge;
-  global.store = new Store(bridge);
+    global.agent = agent;
+    global.bridge = bridge;
+    global.store = new Store(bridge);
+
+    // Reinit is needed to reset the store and the bridge
+    // for data export/import tests where the import should
+    // go to a clean environment.
+    global.reinit = () => {
```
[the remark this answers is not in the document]
sompylasar: Good point.

sompylasar:
> Export+import (in the actual DevTools extension) is broken in this branch. It doesn't runtime error, but no nodes show up in the Profiler after import. Verified that this works on master.

That's sad, I'll look into it. I'm really curious what have we missed. I'm pretty sure it's some `any` somewhere, but why don't the tests show it... Maybe we snapshot the wrong thing.

src/utils.js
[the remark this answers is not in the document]
sompylasar: Okay I understand why this backfired on me. You introduced packing of 3 items into a flat array (which is apparently to designate a nested Map or something?).

```
<<<<<<< HEAD
      const commitDurations = [];
      commitProfilingMetadata.forEach(({ durations }, commitIndex) => {
        for (let i = 0; i < durations.length; i += 3) {
          if (durations[i] === fiberID) {
            commitDurations.push(commitIndex, durations[i + 2]);
            break;
          }
=======
      const commitDurationsInterleaved = [];
      commitProfilingMetadata.forEach((commitProfilingData, commitIndex) => {
        const commitDuration = getFromInterleavedArray(
          commitProfilingData.actualDurations,
          fiberID
        );
        if (commitDuration != null) {
          commitDurationsInterleaved.push(commitIndex, commitDuration);
>>>>>>> Cleanup profile export/import data types, add export/import test
```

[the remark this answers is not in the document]
sompylasar: Got it, it's a map of IDs to a pair of numbers (actualDuration and selfDuration), so `Map<number, [number, number]>` or `Map<number, {actualDuration: number, selfDuration: number]>`.
```
metadata.durations.push(fiberID, actualDuration, selfDuration);
```

src/devtools/views/Profiler/utils.js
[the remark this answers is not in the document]
sompylasar: And I understand why this backfired: this type now needs to change, alongside with the documentation.
```
export type ExportedProfilingDataCommitDetails = Array<{|
  // `actualDurations` map is packed into an interleaved array: key at [i], value at [i + 1].
  actualDurations: Array<number>,
  commitIndex: number,
  interactions: Array<InteractionFrontend>,
  rootID: number,
|}>;
```

src/devtools/ProfilingCache.js
```diff
@@ -225,13 +243,14 @@ export default class ProfilingCache {
...
-            this._pendingProfileSummaryMap.set(
-              profilingSummary.rootID,
-              resolve
-            );
-            this.onProfileSummary(profilingSummary);
+          const profilingSummaryFrontend: ProfilingSummaryFrontend =
+            importedProfilingData.profilingSummary;
+          if (
+            profilingSummaryFrontend != null &&
+            profilingSummaryFrontend.rootID === rootID
```
sompylasar: `selectedCommitIndex` is `null` here:
```
function CommitFlamegraphAutoSizer(_) {
  const _useContext = Object(react__WEBPACK_IMPORTED_MODULE_0__["useContext"])(_context__WEBPACK_IMPORTED_MODULE_8__["StoreContext"]),
        profilingCache = _useContext.profilingCache;

  const _useContext2 = Object(react__WEBPACK_IMPORTED_MODULE_0__["useContext"])(_ProfilerContext__WEBPACK_IMPORTED_MODULE_3__["ProfilerContext"]),
        rendererID = _useContext2.rendererID,
        rootID = _useContext2.rootID,
        selectedCommitIndex = _useContext2.selectedCommitIndex,
        selectFiber = _useContext2.selectFiber;

  const deselectCurrentFiber = Object(react__WEBPACK_IMPORTED_MODULE_0__["useCallback"])(event => {
    event.stopPropagation();
    selectFiber(null, null);
  }, [selectFiber]);
  const profilingSummary = profilingCache.ProfilingSummary.read({
    rendererID: rendererID,
    rootID: rootID
  });
  let commitDetails = null;
  let commitTree = null;
  let chartData = null;

  if (selectedCommitIndex !== null) {
    commitDetails = profilingCache.CommitDetails.read({
      commitIndex: selectedCommitIndex,
      rendererID: rendererID,
      rootID: rootID
    });
    commitTree = profilingCache.getCommitTree({
      commitIndex: selectedCommitIndex,
      profilingSummary
    });
    chartData = profilingCache.getFlamegraphChartData({
      commitDetails,
      commitIndex: selectedCommitIndex,
      commitTree
    });
  }

  if (commitDetails != null && commitTree != null && chartData != null && chartData.depth > 0) {
```

Looks like `ProfilingSummary` cache is caching forever for `rendererID === null` and `rootID === null` (`const key = hashInput(input);` is `null`) which are the values for when a file is imported.

```
function SnapshotSelector(_) {
  const _useContext = Object(react__WEBPACK_IMPORTED_MODULE_0__["useContext"])(_ProfilerContext__WEBPACK_IMPORTED_MODULE_3__["ProfilerContext"]),
        isCommitFilterEnabled = _useContext.isCommitFilterEnabled,
        minCommitDuration = _useContext.minCommitDuration,
        rendererID = _useContext.rendererID,
        rootID = _useContext.rootID,
        selectedCommitIndex = _useContext.selectedCommitIndex,
        selectCommitIndex = _useContext.selectCommitIndex;

  const _useContext2 = Object(react__WEBPACK_IMPORTED_MODULE_0__["useContext"])(_context__WEBPACK_IMPORTED_MODULE_6__["StoreContext"]),
        profilingCache = _useContext2.profilingCache;

  const _profilingCache$Profi = profilingCache.ProfilingSummary.read({
    rendererID: rendererID,
    rootID: rootID
  }),
```

And this `profilingSummaryFrontend.rootID === rootID` check is always false, so it resolves to default empty values:
```
    _defineProperty(this, "ProfilingSummary", Object(_cache__WEBPACK_IMPORTED_MODULE_0__["createResource"])(({
      rendererID,
      rootID
    }) => {
      return new Promise(resolve => {
        const importedProfilingData = this._store.importedProfilingData;

        if (importedProfilingData !== null) {
          const profilingSummaryFrontend = importedProfilingData.profilingSummary;

          if (profilingSummaryFrontend != null && profilingSummaryFrontend.rootID === rootID) {
            this._pendingProfileSummaryMap.delete(rootID);

            resolve(profilingSummaryFrontend);
            return;
```

When I added this check, I assumed that `rootID` is present when we import data, but it may be `null`, whereas `profilingSummaryFrontend.rootID` is always present.

Also we probably cannot use `profilingSummaryFrontend.rootID` from the imported data in the `resolve` as it theoretically has nothing to do with the currently loaded app's `rootID`s (if we load the profile while sitting on another app's page). But maybe we assume that if you load a profile, then it's from the same app that you're sitting at. Anyway, that `rootID` doesn't match the currently selected `rootID` so most likely useless and even confusing downstream.

sompylasar: After I removed this check, the DevTools started showing an error `An error was thrown: "Could not find root node with id "null" in commit tree"`.

sompylasar: OK fixed, missed one more check.

Also there was this type mismatch (`resolve` accepts `FiberCommitsBackend` but was given `FiberCommitsFrontend`, and thanks to Flow's non-nominal typing, this was ignored because they are structurally compatible):
```diff
-          this.onFiberCommits(fiberCommitsFrontend);
+          this._pendingFiberCommitsMap.delete(pendingKey);
+          resolve(fiberCommitsFrontend);
```

There were also missing cache invalidations:
```diff
      this._pendingCommitDetailsMap.clear();
+     this._pendingFiberCommitsMap.clear();
+     this._pendingInteractionsMap.clear();
      this._pendingProfileSummaryMap.clear();
```

src/__tests__/profiling-test.js
```diff
@@ -16,17 +16,19 @@ describe('profiling', () => {
...
         store.profilingOperations,
         store.profilingSnapshots,
         rootID,
```
sompylasar: There's a bug hiding here because we do not use the help that Flow opaque types can provide:
```
        const exportProfilingSummary = prepareExportedProfilingSummary(
          store.profilingOperations,
          store.profilingSnapshots,
          rootID,
          rendererID
        );
```
```
export const prepareExportedProfilingSummary = (
  profilingOperations: ImportedProfilingDataProfilingOperations,
  profilingSnapshots: ImportedProfilingDataProfilingSnapshots,
  rendererID: number,
  rootID: number
) => {
```
All numbers are the same, but not really.

sompylasar:
I moved `exportImportHelper` to `src/__tests__/utils.js` as `utils.exportImportHelper(bridge, store, rendererID, rootID)`, in case you want to use it in the more detailed `src/__tests__/profilingCharts-test.js` and `src/__tests__/profilingCommitTreeBuilder-test.js`.

sompylasar:
I merged your changes from `cleanup-profile-data-types-continued` into here, looks like there weren't many. Please have a look.

src/__tests__/setupTests.js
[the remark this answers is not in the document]
sompylasar: I decided to keep this function to ensure this pattern of init/destroy is preserved, feel free to push me to remove it if you don't want to keep dead code around until it's actually used.

src/backend/types.js
```diff
@@ -123,7 +123,7 @@ export type InteractionBackend = {|
 
 export type CommitDetailsBackend = {|
   commitIndex: number,
-  // Tuple of id, actual duration, and (computed) self duration
+  // An interleaved array: fiberID at [i], actualDuration at [i + 1], computed selfDuration at [i + 2].
```
sompylasar: I hope you don't mind rephrasing the comment like this. The indexes are deliberately written as they are in code so that you can search for `[i + 1]` pattern and find them. Unfortunately `.push(one, two, three)` doesn't use the indexes.

src/__tests__/setupTests.js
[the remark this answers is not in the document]
bvaughn: If we decide to add a test that needs re-init, I think we should just export the init function (rather than attach it to the global).

bvaughn:
Thanks for iterating on this with me!

src/devtools/ProfilingCache.js
```diff
@@ -123,6 +124,7 @@ export default class ProfilingCache {
         }
 
         // If no profiling data was recorded for this root, skip the round trip.
+        this._pendingCommitDetailsMap.delete(pendingKey);
```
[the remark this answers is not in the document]
sompylasar: Just a precaution I guess. Defensive programming.

src/devtools/ProfilingCache.js
```diff
@@ -143,29 +145,27 @@ export default class ProfilingCache {
...
+            if (selfDuration != null) {
+              commitDurations.push(commitIndex, selfDuration);
+            }
+          });
+          const fiberCommitsFrontend: FiberCommitsFrontend = {
+            commitDurations,
+            fiberID,
+            rootID,
+          };
+          this._pendingFiberCommitsMap.delete(pendingKey);
```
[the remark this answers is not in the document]
sompylasar: Defensive programming. Delete then resolve, to make sure nothing is leaking.

src/backend/renderer.js
```diff
@@ -772,25 +773,25 @@ export function attach(
...
   }
 
   function recordMount(fiber: Fiber, parentFiber: Fiber | null) {
     const isRoot = fiber.tag === HostRoot;
-    const id = getFiberID(getPrimaryFiber(fiber));
+    const fiberID = getFiberID(getPrimaryFiber(fiber));
```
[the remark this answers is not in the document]
sompylasar: Yeah I agree. I made them communicate their type clearer in the parts of code that I read, but most likely I missed other places. ParentID and OwnerID are still way more context-independent names than just ID.

PR #250, opened 2019-05-01
Clear profiling data for root when the root is unmounted. Fix profiling screenshots data structure to map rootID to commitIndex.

### LINKED ISSUE #232: Clear profiling data for root when the root is unmounted

sompylasar:
Fix profiling screenshots data structure to map rootID to commitIndex

Propagate `rootID` throughout the code for `captureScreenshot`.

Rename private profiling maps of `store` to make relations more clear.

Fix missing cleanup for screenshots data in `set importedProfilingData` of `store`.

Clear profiling data for root when the root is unmounted

Fixes #232

### DIFF
--- a/src/__tests__/profiler-test.js
+++ b/src/__tests__/profiler-test.js
@@ -1,0 +1,60 @@
+// @flow
+
+describe('Profiler', () => {
+  let React;
+  let ReactDOM;
+  let TestUtils;
+  let store;
+
+  const act = (callback: Function) => {
+    TestUtils.act(() => {
+      callback();
+    });
+    jest.runAllTimers(); // Flush Bridge operations
+  };
+
+  beforeEach(() => {
+    store = global.store;
+
+    React = require('react');
+    ReactDOM = require('react-dom');
+    TestUtils = require('react-dom/test-utils');
+  });
+
+  it('should start and stop profiling, handle root unmounting', async () => {
+    const Parent = ({ count }) =>
+      new Array(count).fill(true).map((_, index) => <Child key={index} />);
+    const Child = () => <div>Hi!</div>;
+
+    const containerA = document.createElement('div');
+    const containerB = document.createElement('div');
+
+    act(() => {
+      ReactDOM.render(<Parent key="A" count={3} />, containerA);
+      ReactDOM.render(<Parent key="B" count={2} />, containerB);
+    });
+    expect(store).toMatchSnapshot('1: mount');
+
+    act(() => {
+      store.startProfiling();
+    });
+    expect(store).toMatchSnapshot('2: profiling started');
+
+    act(() => {
+      ReactDOM.render(<Parent key="A" count={4} />, containerA);
+      ReactDOM.render(<Parent key="B" count={1} />, containerB);
+    });
+    expect(store).toMatchSnapshot('3: update');
+
+    act(() => ReactDOM.unmountComponentAtNode(containerB));
+    expect(store).toMatchSnapshot('4: unmount B');
+
+    act(() => ReactDOM.unmountComponentAtNode(containerA));
+    expect(store).toMatchSnapshot('5: unmount A');
+
+    act(() => {
+      store.stopProfiling();
+    });
+    expect(store).toMatchSnapshot('6: profiling stopped');
+  });
+});
--- a/src/devtools/store.js
+++ b/src/devtools/store.js
@@ -185,27 +185,31 @@ export default class Store extends EventEmitter {
 
   assertEmptyMaps() {
     // This is only used in tests to avoid memory leaks.
-    if (this._idToElement.size !== 0) {
-      throw new Error('Expected _idToElement to be empty.');
-    }
-    if (this._ownersMap.size !== 0) {
-      throw new Error('Expected _ownersMap to be empty.');
-    }
-    if (this._profilingOperationsByRootID.size !== 0) {
-      throw new Error('Expected _profilingOperationsByRootID to be empty.');
-    }
-    if (this._profilingScreenshotsByRootID.size !== 0) {
-      throw new Error('Expected _profilingScreenshotsByRootID to be empty.');
-    }
-    if (this._profilingSnapshotsByElementID.size !== 0) {
-      throw new Error('Expected _profilingSnapshotsByElementID to be empty.');
-    }
-    if (this._rootIDToCapabilities.size !== 0) {
-      throw new Error('Expected _rootIDToCapabilities to be empty.');
-    }
-    if (this._rootIDToRendererID.size !== 0) {
-      throw new Error('Expected _rootIDToRendererID to be empty.');
-    }
+    const assertOneMap = (mapName, map) => {
+      if (map.size !== 0) {
+        throw new Error(
+          `Expected ${mapName} to be empty, got ${
+            map.size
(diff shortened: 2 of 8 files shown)


sompylasar:
As of now, there's still some weird leftovers in some of the maps. I'd appreciate if you @bvaughn could take a look where do these `11` and `12` come from. I tried but I didn't understand what was going on.

```
  ● Profiler › should start and stop profiling, handle root unmounting

    Expected _profilingOperationsByRootID to be empty, got 2: Store {
      _events: {},
      _eventsCount: 0,
      _maxListeners: undefined,
      _captureScreenshots: false,
      _collapseNodesByDefault: true,
      _hasOwnerMetadata: false,
      _idToElement: Map {},
      _importedProfilingData: null,
      _isProfiling: true,
      _ownersMap: Map {},
      _profilingOperationsByRootID:
       Map {
         11 => [ Uint32Array [ 1, 11, 0, 2, 2, 8, 7, 2, 1, 6 ] ],
         12 => [ Uint32Array [ 1, 12, 0, 2, 5, 10, 5, 4, 3, 2, 2, 1, 1 ] ] },
      _profilingScreenshotsByRootID: Map {},
      _profilingSnapshotsByElementID: Map {},
      _revision: 6,
      _roots: [],
      _rootIDToCapabilities: Map {},
      _rootIDToRendererID: Map {},
```

src/devtools/store.js
```diff
@@ -870,6 +889,11 @@ export default class Store extends EventEmitter {
               throw new Error(`Node ${id} was removed before its children.`);
             }
 
+            // The following call depends on `getElementByID`
+            // which depends on the element being in `_idToElement`,
+            // so we have to do it before removing the element from `_idToElement`.
+            this._clearProfilingSnapshotRecursive(id);
```
bvaughn: In a meeting so I'm a bit distracted, but I think this change would break things.

It's okay for us to clear _root_ data, because if you can't select the root (in the Components tree) then there's no way to view the Profiler charts to begin with, but it's _not_ safe to clear the Element data, because we need these initial snapshots to render the charts in the flame graph for elements that existed during some commits, but not during other commits.

So basically, I think this should be moved down into the `if (parentID === 0) {` block below (so that it's only done for roots).

sompylasar: > It's okay for us to clear root data, because if you can't select the root (in the Components tree) then there's no way to view the Profiler charts to begin with, but it's not safe to clear the Element data, because we need these initial snapshots to render the charts in the flame graph for elements that existed during some commits, but not during other commits.

Sure, that's why we review code, I thought about this, too, and it was under the "is root?" condition initially, I just thought we might be leaking something if we don't. But now I think you're right, they should be cleaned up when their root cleans up recursively.

src/__tests__/profiler-test.js
```diff
@@ -0,0 +1,60 @@
...
+    act(() => ReactDOM.unmountComponentAtNode(containerB));
+    expect(store).toMatchSnapshot('4: unmount B');
+
+    act(() => ReactDOM.unmountComponentAtNode(containerA));
+    expect(store).toMatchSnapshot('5: unmount A');
+
+    act(() => {
+      store.stopProfiling();
+    });
+    expect(store).toMatchSnapshot('6: profiling stopped');
```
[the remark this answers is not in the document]
sompylasar: > Not sure how useful the snapshots are in this specific test. Seems like the thing we're testing is more than isProfiling starts and stops, and that data is cleared after the root is unmounted- but the test doesn't assert anything about that.

For sure, I didn't spend much time on making useful snapshots, this was mechanical copy-paste.

src/devtools/store.js
```diff
@@ -181,27 +185,31 @@ export default class Store extends EventEmitter {
...
-    if (this._profilingSnapshot.size !== 0) {
-      throw new Error('Expected _profilingSnapshot to be empty.');
-    }
-    if (this._rootIDToCapabilities.size !== 0) {
-      throw new Error('Expected _rootIDToCapabilities to be empty.');
-    }
-    if (this._rootIDToRendererID.size !== 0) {
-      throw new Error('Expected _rootIDToRendererID to be empty.');
-    }
+    const assertOneMap = (mapName, map) => {
```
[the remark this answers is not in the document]
sompylasar: Well it's local and the emphasis is to what distinguishes it from `assertEmptyMaps` (plural). 🙂 I usually prefer not to reorder words so I'd rather `assertEmptyMap` or rename both: `assertAllMapsEmpty`, and `assertOneMapEmpty` or `assertSingleMapEmpty`. Bikeshedding FTW.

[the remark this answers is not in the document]
bvaughn: Admittedly, this is just a nit 😄  but I find `assertOneMap` to be a confusing name.  When I first read it, I thought "is this asserting that the map contains a single element?" which obviously, I quickly saw that it didn't.

But yeah maybe `assertEmptyMap`

bvaughn:
Want me to finish this out and land it? Or should I hold off on you to push an update?

sompylasar:
Please finish if you have time. Thanks!

bvaughn:
Will do.

src/devtools/store.js
bvaughn: So this _would_ have caused a runtime error, bug I think fixing it is a bit trickier than I assumed because of the fact that we unmount children before their parents. (This means that although the only time it's safe to clear this map is when a root is unmounted, it's also _too late_ to recursively clear it at the point.) 😁

Need to give this a little more thought.

bvaughn: Looks like another source of complication is when we remove a root, the renderer interface's `flushPendingEvents()` gets a new ID for that root (since we have already cleared out the `fiberToIDMap`) and this causes the `Store` to add a new entry to the `_profilingOperationsByRootID` map 😁

sompylasar: Awesome, thanks for looking, looks like this surfaced deeper problems with the implementation.

bvaughn: I don't know if I'd categorize it that way 😁 we just weren't trying to reset on unmount before

sompylasar: Yeah yeah YAGNI approach. I hate it, if you ask me, and prefer to think several steps forward at least in such memory/resource management scenarios where there's some alloc or add something, I believe there must be mirroring dealloc or remove, most of the times, otherwise you're shooting in the foot.

bvaughn: I think it would be silly to suggest that either of us aren't thinking a few steps ahead 😄 We're just prioritizing differently.

sompylasar: Agree. 😄 I do prioritize correct resource management as I've seen numerous examples of non-linear increase in bugs and effort to maintain the code where this is not addressed timely, if not built with this requirement initially.

PR #251, opened 2019-05-01
Use WebpackDevServer for local testing

gaearon:
Three purposes:

* This ensures first load doesn't serve a stale build from previous development session. I always found it a bit confusing.
* We can live edit CSS now:

* Selfish goal: this will make it easier for me to test and dogfood the new React hot reloading implementation I'm experimenting with. (This PR doesn't enable it but it's useful to me locally.)

### DIFF
--- a/shells/dev/src/devtools.js
+++ b/shells/dev/src/devtools.js
@@ -38,7 +38,7 @@ mountButton.addEventListener('click', function() {
   }
 });
 
-inject('./build/app.js', () => {
+inject('./app.js', () => {
   initDevTools({
     connect(cb) {
       const bridge = new Bridge({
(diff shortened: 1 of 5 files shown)


shells/dev/src/devtools.js
```diff
@@ -78,7 +78,10 @@ inject('./build/app.js', () => {
 
         // Initialize the backend only once the DevTools frontend Store has been initialized.
         // Otherwise the Store may miss important initial tree op codes.
-        inject('./build/backend.js');
+        inject('./backend.js', () => {
+          // Clear noisy webpack "WDS: ..." output.
+          console.clear();
```
gaearon: This is gross but webpack is hopeless. I've spent an hour trying to get it to shut up with these logs, and I didn't reach success. It seems like some race condition is preventing the silent config from kicking in, so I just clear all initial logs. Could be annoying if you debug the init path but I figured it's not too bad and it's always clearly shown in DevTools that console was cleared — so you can always comment it out.

bvaughn:
Ack  I think this broke the  deploy  script. react-devtools-experimental.now.sh is now blank because it's  trying to load a file that doesn't exist

bvaughn:
I did a local revert and deployed the demo site to buy  us some  time. I'll file a follow up issue

PR #258, opened 2019-05-03
Experimenting with profiler tests

### LINKED ISSUE #252: Add unit tests for Profiler
Even if we don't test the Profiler graphs, we can still test:
- [x] (#258) Renderer interface collecting profiling data
- [x] (#258) `ProfilerCache` returning promises that resolve to the collected data
- [x] (#258) Various chart builds generate reasonable data

bvaughn:
Resolves #252

Adds test coverage for profiling.

Note that this coverage does not include the SVG charts, but does include everything else (store, renderer interface, suspense cache, chart builders).


bvaughn:
Have a full set of initial tests written, but they're blocked by https://github.com/facebook/react/pull/15567

bvaughn:
This PR is still blocked on a canary (https://github.com/facebook/react/pull/15567) but I've verified that it passes once that fix is in place.

src/__tests__/profiling-test.js
```diff
@@ -0,0 +1,266 @@
+// @flow
+
+describe('profiling', () => {
+  let React;
+  let ReactDOM;
+  let Scheduler;
+  let SchedulerTracing;
+  let TestRenderer;
+  let store;
```
sompylasar: Off-topic but in https://github.com/bvaughn/react-devtools-experimental/pull/248 I found helpful for writing tests to annotate our globals with typehints (like `: Store`, `: Bridge`). What do you think?

bvaughn: I'm not used to Flow-typing tests because the React repo (and most of my OSS projects) don't do it, but that sounds like a good idea.

bvaughn: I think after 705cd9b, this new test should be pretty covered. Thoughts?

sompylasar: Yeah looks great, you even got a type mismatch spotted :-)
```diff
-   return ids[0];
+   return parseInt(ids[0], 10);
```

bvaughn:
Okay. I updated to use the newly-released react canary which should fix the failing tests.

PR #262, opened 2019-05-08
Profiler: Improve labels for times that round to 0.0ms

submetu:
reactpullrequest

So added the enhancement in this way. 
Maybe you can guide me if there is a better way to do this. 
Also, the function name `fineTune` didn't sit quite well with me. 
If we decide to create a function for handling this, we should probably choose a better name. 

PS: The GIF is extremely slow, sorry about that. I couldn't find a good .mov -> .gif online converter.


submetu:
Yes, that makes sense. 
I didn't think of messing with the `formatDuartion()` function. 

But will do that now.

submetu:
So I edited `formatDuration()` and `formatTime()` methods to return `'<0.1'` if the formatted value is 0.

bvaughn:
Hmm! I'm not sure if `formatTime` should do this. 🤔 I think it makes sense for something to occur at 0s (relative to the start of profiling- in the case of reload-and-profile). It's _unlikely_ to happen, but it isn't _confusing_. But it _is_ confusing to see something that took "0.0 seconds" to do.

I think maybe we should only do this for the duration helper 😄 

Also CI is failing because some snapshot tests are outdated now (some labels changed from "0.0" to "0") but this is fine. I can fix it, or you can if you want to re-run Jest with `-u` to update them.

submetu:
Yes, that is right. 😄 
Time can be 0s. 

Oh no worries 👍 , I updated the snapshots and removed the extra formatting from `formatTime()`

submetu:
👍

submetu:
Hey Brian
Just wondering why my name is not showing up in the contributors list of this project ? 😄 
Is it because of a git username/email mismatch ? 🤔

bvaughn:
I don't know. Maybe GitHub is falling behind on contributor data.

PR #264, opened 2019-05-09
Fetch owners list from renderer (using suspense)

### LINKED ISSUE #260: Component filtering breaks owners tree and list views
- [ ] Owners tree

Entering the owners tree view can cause runtime errors (in `reduceOwnersState`) if any owners have been filtered out of the tree. We need to guard against this.

- [x] 17516b7: "Rendered by" list

It's a bad UX to click on an owner (in the right hand panel) if the owner has been filtered in the tree. Nothing breaks, but the selection just kind of goes blank. The owner probably shouldn't be clickable in that case.

bvaughn:
Owners in the list may have been filtered out of the Store, but in the owners list view- it's important to still show them. The frontend cannot do this on its own, so this list needs to come from the renderer interface.

- [ ] Tests

Resolves #260


bvaughn:
I'm going to leave the two `setState` change for a future optimization. Landing the rest of this as is since it's fixed a couple of small bugs and added lots of tests.

PR #266, opened 2019-05-10
DO NOT MERGE - rewrite test act helpers based on react/#15591

threepointone:
This needs https://github.com/facebook/react/pull/15591 to land to work correctly. 

This simplifies your test helpers to loop until all timers are flushed (including the ones that get queued after updates), and works in concurrent mode.


threepointone:
(I also wanted to rename `actSuspense` to `actAsync` since it isn't exactly suspense related, but I'll leave that to you)

bvaughn:
Since https://github.com/facebook/react/pull/15591 was merged and we're running on a canary release that includes it, I tried applying these changes in master. Looks like everything works _except_ for [`inspectedElementContext-test.js`](https://github.com/bvaughn/react-devtools-experimental/blob/master/src/__tests__/inspectedElementContext-test.js) which _always_ fails with an infinite loop bug here:
https://github.com/bvaughn/react-devtools-experimental/blob/7aaef6f41fa0dc0b13554a2e0800a34187a4ddfc/src/__tests__/utils.js#L24-L29

Not sure yet what would be causing this.

bvaughn:
Let's talk when we're both back in the office on Tuesday, Sunil. In the meanwhile I'll push my work in progress here  https://github.com/bvaughn/react-devtools-experimental/pull/new/threepointone-act-async

threepointone:
boop, have a look and let me know if this solves your problems

threepointone:
sorry, I should be a commie about this 
*our* problems

bvaughn:
Cool, thanks Sunil!

bvaughn:
Hey uh...looks like `jest.getTimerCount()` doesn't exist in the version of Jest used in the React repo. Wondering what you're suggesting folks use who are using a version of Jest older than v24 (like me now, heh heh)

threepointone:
Oh hmm. We could just update to jest 24, if you could help resolve this issue with react-is? https://github.com/facebook/react/pull/15778

threepointone:
(Alternately, there may be a way of getting at the timer count with a hidden api, I’ll have a look)

bvaughn:
Upgrading might be nice

SimenB:
I'm pretty sure there's no way to access a count of scheduled timers in 23 without using a custom `testEnvironment` which could expose the `FakeTimers` instance on the `global` injected into the test

bvaughn:
That doesn't sound out of the question, unless I'm misunderstanding something.

It looks like the [`fake-timers` object exposes a `mockGetTimersCount()`](https://github.com/facebook/jest/blob/770799779fc853ab6285727e8938efd06e694d5f/packages/jest-util/src/fake_timers.js#L126) which I think might satisfy the need my tests have.

Not sure how to get a handle on that object though.

bvaughn:
Oh I think can just import it off of `jest-utils` huh. Let me try that!

bvaughn:
Okay. I think that path would entail:
* Extending the default Jest env (e.g. `jest-environment-jsdom`)
* Specifying a custom `testEnvironment` in my Jest config to point to the subclass
* Getting a handle on `this.fakeTimers` in my environment subclass
* Exposing that handle to tests via some global ref (this is the part I'm least sure about, since I'm not sure about sequencing and any potential concurrency issues)

bvaughn:
Actually it looks like Jest sets `mockGetTimersCount` on the `global` object itself 😆  I don't even need to.

SimenB:
Ah, I forgot about that. Removed in 24 and replaced by `getTimerCount` in https://github.com/facebook/jest/pull/7285. Great you were able to find it

bvaughn:
No problem 😄 Now I'm more familiar with Jest internals, having traced through this, so not wasted effort.

SimenB:
Awesome! To answer your above question (even though they're striked out, so you probably figured it out), all your steps are correct. Should probably `delete` in `teardown` as well to avoid a potential memory leak, but all tests gets their own `global`, so there _shouldn't_ have been any concurrency issues

bvaughn:
You rock. Thansk~

PR #269, opened 2019-05-12
Profiler shows which fibers scheduled an update, and why a given component rendered (do not merge)

bvaughn:
### Note that this PR is blocked by React [PR 15658](https://github.com/facebook/react/pull/15658) and unit tests are expected to fail unless using a version of React built from from the [`scheduler-tracking` branch](https://github.com/bvaughn/react/tree/scheduler-tracking).

I'm exploring how the Profiler UI might help answer a few common questions:

### Why did this commit happen in the first place?

This feature requires a build of react-dom from the [`scheduler-tracking` branch](https://github.com/bvaughn/react/tree/scheduler-tracking).

Here's an example that shows how batched updates will list multiple schedulers:
why-batched

And here's an example that illustrates both suspense (how the scheduler is continued across the ping) _and_ how we could potentially handle a scheduler that is unmounted as a result of the udpate:
why-suspenseKapture 2019-05-15 at 11 17 40

### Why did component X re-render as part of it?

This is an example of how we could show _what changed_ between renders:
whatwhat

Note that hooks are tricky to handle, so for now we don't try to display _what_ hook changed.


src/backend/renderer.js
```diff
@@ -1308,7 +1401,14 @@ export function attach(
             timestamp: interaction.timestamp - profilingStartTime,
           })
         ),
+        updaters:
```
bvaughn: Note to self: We should probably respect filtering preferences here. Otherwise we may end up showing "an unmounted component" when it was really just a filtered-out component.

bvaughn:
I'm going to split off the change tracking from the updater tracking, since the former is not blocked by a React PR.

#310

bvaughn:
Closing all PRs at this repo as part of migrating to github.com/facebook/react

PR #274, opened 2019-05-15
Show commit priority levels in Profiler UI (if available)

bvaughn:
**Note that this PR is blocked by React [PR 15664](https://github.com/facebook/react/pull/15664) and unit tests are expected to fail unless using a version of React built from from the [`devtools-commit-priority-level` branch](https://github.com/bvaughn/react/tree/devtools-commit-priority-level).**

### What was the priority of this commit?

Some commits are high priority. Others are pre-rendering offscreen work. Up until now though, the Profiler has treated them all the same.

Here's an example app:
```js
function PriorityLevels() {
  const [count, setCount] = useState(0);

  const startSequence = useCallback(() => {
    setCount(1);
    next(() => setCount(2));
  }, []);

  return (
    <Fragment>
      <h1>Priority Levels</h1>
      <button onClick={startSequence}>start sequence</button>
      {count >= 1 && <Text>One</Text>}
      {count >= 2 && <Text>Two</Text>}
      {count >= 2 && (
        <div hidden>
          <Text>Three</Text>
        </div>
      )}
    </Fragment>
  );
}

const Text = ({ children }) => children;
```

And here's what the UI could look like:
priority-Kapture 2019-05-15 at 15 21 57


src/backend/types.js
```diff
@@ -32,6 +32,16 @@ export type HookType =
...
 
+// Priority level is copied from React and should be kept in sync:
+// https://github.com/facebook/react/blob/master/packages/react-reconciler/src/SchedulerWithReactIntegration.js
+export opaque type ReactPriorityLevel = 99 | 98 | 97 | 96 | 95 | 90;
+export const ImmediatePriority: ReactPriorityLevel = 99;
+export const UserBlockingPriority: ReactPriorityLevel = 98;
+export const NormalPriority: ReactPriorityLevel = 97;
+export const LowPriority: ReactPriorityLevel = 96;
+export const IdlePriority: ReactPriorityLevel = 95;
+export const NoPriority: ReactPriorityLevel = 90;
```
bvaughn: Note to self: These would need to live in the renderer interface and be assigned based on renderer version. This was just a quick hack. :)

PR #280, opened 2019-05-22
[Flare] Adds inspection of React event components/props

trueadm:
This PR adds support for inspecting event components from the experimental event API. For now, we only display the event component name and the relevant props for their event component. When you select an element in the tree, it will traverse up the fiber tree and collect all relevant event components.

Screen Shot 2019-05-22 at 3 17 17 PM

I wasn't quite sure how to add tests for this work, as the Flare work is behind a flag and thus not available to be used in tests.


trueadm:
CI is failing because of a snapshot test that is no longer needed. I don't see this error when running locally, so I'm not sure how to fix (even after merging master).

bvaughn:
CI error should show what needs to be done:
Screen Shot 2019-05-22 at 11 31 07 AM

You need to run `yarn test -u`

trueadm:
@bvaughn I did already – that's how I updated the snapshots. Doing it again doesn't remove the obsolete snapshot.

bvaughn:
Hm. Odd.

Running a fresh `rm -rf node_modules && yarn install && yarn test` for me on master doesn't show any failures or updates. I wonder what's different about your setup?

bvaughn:
We should probably add at least one new test that covers this functionality too (to `inspectedElementContext-test.js`)

src/backend/renderer.js
```diff
@@ -1803,6 +1807,23 @@ export function attach(
...
+    while (node !== null) {
+      if (node.tag === EventComponent) {
+        if (events === null) {
+          events = [];
+        }
+        const eventComponentInstance = node.stateNode;
+        const currentFiber = eventComponentInstance.currentFiber;
+        events.push({
+          props: eventComponentInstance.props,
+          displayName: currentFiber.type.displayName,
```
bvaughn: Will this always be a (non-empty) string or should we have a fallback (e.g. "Anonymous")?

Also we should probably use `getDisplayNameForFiber(currentFiber)` for consistency here. (That's what we do elsewhere.)

Looks like `getDisplayNameForFiber` doesn't yet handle `EventComponent` types.

trueadm: `displayName` is no longer optional as far as I'm aware, so we should be fine.

trueadm:
> We should probably add at least one new test that covers this functionality too (to inspectedElementContext-test.js)

I know, but I was unsure how to test it, given that this functionality isn't in `react` or `react-dom` packages because of the Flare flag.

src/backend/renderer.js
```diff
@@ -457,6 +457,10 @@ export function attach(
           case EVENT_TARGET_TOUCH_HIT_STRING:
             return 'TouchHitTarget';
           default:
+            const displayName = elementType.displayName;
```
bvaughn: Still not returning a display name for `EventComponent` types.

src/backend/renderer.js
bvaughn: > Also we should probably use `getDisplayNameForFiber(currentFiber)` for consistency here. (That's what we do elsewhere.)

> Looks like `getDisplayNameForFiber` doesn't yet handle EventComponent types.

Doesn't seem like either fo these changes were made 😁

bvaughn: I guess I can just make them while I'm in here.

bvaughn:
Oh well! Still worth pushing forward with to help people who are early-adopting Flare. Let's do it!

src/backend/renderer.js
trueadm: Sorry I didn’t get around to it today. :( thank you for the help!

bvaughn: No problem!

bvaughn:
No problem! I rebased and merged. Again, not sure why GitHub isn't picking up on the fact that these commits have all been merged:
Screen Shot 2019-05-23 at 9 39 58 AM

Oh well~

Issue #282, opened 2019-05-22
Component selector should be Flare compatible


bvaughn:
DevTools currently relies on `preventDefault` and `stopPropagation` to prevent the selector tool from actually clicking on a DOM element:

https://github.com/bvaughn/react-devtools-experimental/blob/eaaeb439b4d80e72892b897d900568f8cf77ca9c/src/backend/agent.js#L484-L490

Flare does not work with this, so you can't select a component without also potentially changing it in a significant way. We should find a way to fix this!

cc @trueadm

trueadm:
cc @necolas

trueadm:
With Flare, we could provide a prop to <Press> that always does preventDefault and stopImmediatePropagation. Would that be a solution?

bvaughn:
Seems like it could work, yeah. So long as we could identify when to use the Flare approach vs the normal/synthetic event approach.

trueadm:
Maybe I have got this wrong though. Is this using Flare from within dev tools, or when dev tools interacts with a page using Flare? If the later, then my point above won’t work. Let me think about how to do that.

bvaughn:
Check out the code snippet I linked to in the bug description.

Essentially, DevTools needs a way to identify that the user has clicked on a DOM element without the event slipping through to the element itself.

We're already using capture:
https://github.com/bvaughn/react-devtools-experimental/blob/eaaeb439b4d80e72892b897d900568f8cf77ca9c/src/backend/agent.js#L375-L378

trueadm:
Ah, maybe it’s as simple as using pointer events here instead of mouse events? Given the only browser that doesn’t support pointer events is Safari, this should be fine as dev tools isn’t supported on safari.

trueadm:
@bvaughn Actually, I just checked. This should be fine – as we're attaching event listeners in dev tools to the `window`, whilst Flare attaches them to the `document`. On closer inspection, `window` events always fire first, so I'm not sure why we're seeing issues here, other than maybe you're using `stopPropagation`. If you try using `stopImmediatePropagation()` does that fix this issue?

bvaughn:
> If you try using `stopImmediatePropagation()` does that fix this issue?

Will check (when I get the chance) and report back.

trueadm:
I've updated my reply but that should work. :)

bvaughn:
Thansk!

PR #283, opened 2019-05-22
[Flare] Ensure DOM inspection is Flare compatible

trueadm:
This should address the issues in https://github.com/bvaughn/react-devtools-experimental/issues/282. I haven't had the chance to check manually as it's late here; I can check this next week once I'm back from PTO :)

The general solution should work because Flare uses Pointer Events and Dev Tools only uses mouse events. Pointer events fire _before_ mouse events, and also Flare fires many events to do with pressing in both the `pointerdown` and `pointerup` events – so if we `stopPropagation()` in these events, it should block them from propagating to Flare's event listeners on the `document` (dev tools uses the `window`).

I moved the logic for dev tools into the relevant pointer events too – which should be fine as the only major browser that doesn't yet support Pointer Events in Safari and I don't believe we have any extension for dev tools for that platform. I changed the mouse events to only block events (`stopPropagation`, `preventDefault`) from firing on the heritage React event system.

### DIFF
--- a/src/backend/agent.js
+++ b/src/backend/agent.js
@@ -472,7 +472,9 @@ export default class Agent extends EventEmitter {
     window.addEventListener('click', this._onClick, true);
     window.addEventListener('mousedown', this._onMouseDown, true);
     window.addEventListener('mouseup', this._onMouseUp, true);
-    window.addEventListener('mouseover', this._onMouseOver, true);
+    window.addEventListener('pointerover', this._onPointerOver, true);
+    window.addEventListener('pointerdown', this._onPointerDown, true);
+    window.addEventListener('pointerup', this._onPointerUp, true);
   };
 
   startProfiling = () => {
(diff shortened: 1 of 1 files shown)


src/backend/agent.js
```diff
@@ -492,7 +494,9 @@ export default class Agent extends EventEmitter {
     window.removeEventListener('click', this._onClick, true);
     window.removeEventListener('mousedown', this._onMouseDown, true);
     window.removeEventListener('mouseup', this._onMouseUp, true);
-    window.removeEventListener('mouseover', this._onMouseOver, true);
```
bvaughn: Should we still block "mouseover" events like we do for the other mouse/click events?

src/backend/agent.js
```diff
@@ -492,7 +494,9 @@ export default class Agent extends EventEmitter {
     window.removeEventListener('click', this._onClick, true);
```
bvaughn: Should we still be using "click" for the signal to stop inspecting? Or should we move that to pointer down as well?

bvaughn: Ah, I guess we can't cleanup in pointer up or we wouldn't block the "click" huh.

bvaughn:
Seems to fix the problem. Thanks Dominic!

bvaughn:
Rebased and merged (not sure why it didn't auto-close)

src/backend/agent.js
trueadm: Yep, we need to keep "click".

src/backend/agent.js
trueadm: Good point, added that back in, plus `mouseenter` and `pointerenter` to capture other cases.

trueadm:
I've updated the PR before you merged. Not sure if you want those extra changes or not. Feel free to add them in another PR if you do.

PR #310, opened 2019-06-07
Add profiling option to track why a component rendered

bvaughn:
I'm exploring how the Profiler UI might help answer the question: Why did <some component> re-render?

### New preference added
Profiler change tracking preference

### A component mounted
A component was mounted

### A component updated
A component's props and state changed

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -649,7 +649,7 @@ export function attach(
   }
 
   function getChangeDescription(
-    prevFiber: Fiber,
+    prevFiber: Fiber | null,
     nextFiber: Fiber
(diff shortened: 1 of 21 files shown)


src/backend/renderer.js
```diff
@@ -350,7 +352,7 @@ export function attach(
   // The unmount operations are already significantly smaller than mount opreations though.
   // This is something to keep in mind for later.
   function updateComponentFilters(componentFilters: Array<ComponentFilter>) {
-    if (this._isProfiling) {
+    if (isProfiling) {
```
bvaughn: This should not have slipped through before 😦

Issue #313, opened 2019-06-10
State Not Visible in React 15.6.1 Application


agreene-coursera:
I was trying out the React 15 compatible release on our logged out homepage (https://coursera.org) and noticed that none of our components' state or legacy context was being displayed. The top `SearchBar` is a stateful class component with context that you can use to reproduce.

![image](https://user-images.githubusercontent.com/30501355/59226207-51c19e00-8b87-11e9-95f2-23101c30fc09.png)

I also noticed a ~a dozen new errors in the console from what looks like a devtools file that may help with debugging. 

Example Console Error:
![image](https://user-images.githubusercontent.com/30501355/59226349-a82edc80-8b87-11e9-91b6-3cb96209d5f4.png)

The besides this issue, the new devtools look great and I'm psyched to get to start using them 😄

bvaughn:
Thanks for the report!

cc @gaearon

bvaughn:
Thanks for the bug report @agreene-coursera! Just deployed a fix (hopefully) to DevTools. You can pull it in by [re-installing](https://github.com/bvaughn/react-devtools-experimental#installation).

agreene-coursera:
Happy to help, the updated version works like a charm!

PR #319, opened 2019-06-12
Fix some legacy backend bugs

gaearon:
Fixes https://github.com/bvaughn/react-devtools-experimental/issues/313.

- Ignore SSR "fake mounts" (meaning client-side renderToString calls). Those happen to be seen by DevTools but won't work and should be ignored.
- Don't forget to associate components with roots when crawling a tree later.
- Read state and context from the right object.
- Distinguish classes and functions, and put the whole class instance into `$r` like before.


gaearon:
I didn't add any tests because I'm a bad person

PR #326, opened 2019-06-15
fix: ensure when-rendered content is never hidden

### LINKED ISSUE #324: "Why did this render?" section's CSS hides the reason when there are lots of renders
When there's a very long list of renders, the "Why did this render?" section is collapsed so that it hides the reason (and sometimes even hides the label!). When I changed the CSS of that section from `overflow-y:auto` to `overflow-y:visible`, then the reason always shows up correctly regardless of how many renders are in the list below it. 

When dev tools are docked on the bottom (label partially hidden):
image

When dev tools are docked on the right (reason hidden):
image

**Repro steps:** 

1. `git clone https://github.com/justingrant/why-render-repro.git`
2. `cd why-render-repro`
3. `npm install`
4. `npm start`
5. Open React Dev Tools, go to the profiler tab, enable the "why render" setting, and start profiling
6. Scroll the calendar up ~10 rows and down ~10 rows (enough to cause a lot of renders)
7. Stop profiling
8. Click on the List component in the render tree

**Expected**: Reason shown under "Why did this render?"
**Actual**: "Why did this render?" apparently shown with blank reason... until you scroll the "Why did this render" label and then you'll find the reason was hiding under the "Rendered at:" heading but was hidden by the overflow-y: auto. If you change it to overflow-y: visible then the reason shows up normally.

justingrant:
Fixes #324.  This PR ensures that short sections of content in the profiler sidebar are always shown, even when there are a large number of renders shown in the sidebar. Without this PR, if there are many renders then the CSS `overflow-y: auto` would hide content like "why did this render?" until you scrolled down to see it, which is undiscoverable because there's no scrollbar or other indicator letting you know there's hidden content.

This PR divides sidebar content into "short" content (always show all of it without scrolling) and "maybe long" content (scrolling is OK). The "short" content gets `overflow-y: visible` while the "maybe long" content gets `overflow-y: auto`.

I tested locally in latest Chrome and latest Firefox and PR works as expected. 

For verification steps, see #324.

### DIFF
--- a/src/devtools/views/Profiler/SidebarSelectedFiberInfo.css
+++ b/src/devtools/views/Profiler/SidebarSelectedFiberInfo.css
@@ -10,6 +10,13 @@
 .Content {
   padding: 0.5rem;
   user-select: none;
+}
+
+.ContentShort {
+  overflow-y: visible;
+}
+
+.ContentMaybeLong {
   overflow-y: auto;
 }
 
--- a/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
+++ b/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
@@ -74,7 +74,7 @@ export default function SidebarSelectedFiberInfo(_: Props) {
         profilerStore={profilerStore}
         rootID={((rootID: any): number)}
       />
-      <div className={styles.Content}>
+      <div className={`${styles.Content} ${styles.ContentMaybeLong}`}>
         {listItems.length > 0 && (
           <Fragment>
             <label className={styles.Label}>Rendered at</label>: {listItems}
@@ -116,7 +116,7 @@ function WhatChanged({
 
   if (changeDescription.isFirstMount) {
     return (
-      <div className={styles.Content}>
+      <div className={`${styles.Content} ${styles.ContentShort}`}>
         <label className={styles.Label}>Why did this render?</label>
         <div className={styles.WhatChangedItem}>
           This is the first time the component rendered.
@@ -199,7 +199,7 @@ function WhatChanged({
   }
 
   return (
-    <div className={styles.Content}>
+    <div className={`${styles.Content} ${styles.ContentShort}`}>
       <label className={styles.Label}>Why did this render?</label>
       {changes}
     </div>


src/devtools/views/Profiler/SidebarSelectedFiberInfo.css
```diff
@@ -10,6 +10,13 @@
 .Content {
   padding: 0.5rem;
   user-select: none;
+}
+
+.ContentShort {
+  overflow-y: visible;
```
bvaughn: Hm...I don't think "overflow: visible" feels like the right solution here. I don't think we ever want this content to render outside of the padding box. I _think_ `flex-shrink: 0` might be a more appropriate fix here for each of the content sections, and then `overflow-y: auto` on the parent container (although not including the header bit). Thoughts?

justingrant: @bvaughn - unfortunately my knowledge of post-2010 CSS is pretty spotty so I'm the wrong guy to ask about whether `flex-shrink` would work here. I could try it out, but I'd worry that I might miss some cases.

Playing devil's advocate: the "short content" is always on top of the sidebar, and the sidebar is already clipped by the bottom of the browser frame. Even if the short content extends lower than the bottom, it won't affect what users see.  So even if using flex might be more correct, I'm not sure it's worth the effort until you start putting other can't-occlude content at the bottom of the sidebar. What do you think?

bvaughn: I think the "correct" behavior is for the whole bottom area (excluding the title bar) to scroll as one, rather than individual `.Content` panels scrolling (since these don't _look_ like separate, scrollable panels- no border, etc)

I went ahead and made this change prior to merging your diff 😄 If you'd like, you can re-install and give it a spin and see how it feels.

justingrant: Haha, OK I'll try it out. BTW, this "why did this render?" feature is a winner. Debugging my toughest issues is much faster now. Also, beyond debugging, it's helped me learn about how React works-- I feel like I've learned more about how React works in the last few days than the previous few months. Thanks so much for building this!

bvaughn: Nice! 😁 Thanks!

PR #331, opened 2019-06-19
[Flare] Update event component displayName

trueadm:
This is for https://github.com/facebook/react/pull/15927 (if it lands). It updates the `displayName` location on event components.

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -479,7 +479,7 @@ export function attach(
       case IndeterminateComponent:
         return getDisplayName(resolvedType);
       case EventComponent:
-        return type.displayName || 'EventComponent';
+        return type.responder.displayName || 'EventComponent';
       case EventTarget:
         switch (getTypeSymbol(elementType.type)) {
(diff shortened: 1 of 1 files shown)


src/backend/renderer.js
```diff
@@ -479,7 +479,7 @@ export function attach(
       case IndeterminateComponent:
         return getDisplayName(resolvedType);
       case EventComponent:
-        return type.displayName || 'EventComponent';
+        return type.responder.displayName || 'EventComponent';
```
bvaughn: Will `type` always have a `responder` value or should we add a null check here? (Will this break FB5 if the DT sync lands before React?)

trueadm: It will always have a responder.

trueadm:
This might not be needed, as I plan on possibly changing the signature again in React core to allow for a different structure than can be minified better.

bvaughn:
lol

Your comment has been edited from:
> This might not be needed, as I plan on possibly changing the signature again in React core to allow for a different structure than can be minified better.

To:
> It would be good to get this in sometime :)

PR #335, opened 2019-06-22
Standalone NPM packages and React Native support

### LINKED ISSUE #334: React Native: Standalone package and basic RN support
Need to create `react-devtools` and `react-devtools-core` packages to support React Native, Safari, iframes, and other embedded cases like Nuclide and Flipper.
- [x] Test standalone shell with Safari
- [x] Test standalone shell with iOS React Native (with and without v8)
- [ ] Test standalone shell with Android React Native
- [ ] Test embedded integration with [Flipper](https://github.com/facebook/flipper)

Some features will require changes to support the RN environment:
- [x] Component filters will need to be saved somewhere between reloads. (Perhaps they will need to be sent eagerly on connect, similar to how we do for reload-and-profile in the browser.)
- [x] Inspecting an element in React Native should also select it in DevTools.
- [x] ~~The selection tool in DevTools should enable clicking on a native view in React Native to select its owner.~~
- [x] Overlay/highlight component to show the currently-selected component. (Currently this is just disabled for RN.)
- [x] ~~Determine if it would be possible to support reload-and-profile functionality for React Native.~~
- [x] ~~Don't hide the DevTools app if Profiler has data when an app unmounts.~~
- [x] React Native style editor plug-in. (Maybe we can use  view config as a whitelist of all valid prop names?)
- [x] ~~Fabric support (stretch goal)~~
- [x] ~~Special tag/badge for native components (React Native will be dropping the "RCT" prefix soon).~~
- [x] Inspect component in IDE (launch editor) feature.

In addition, because the v4 backend API is incompatible with v3, releasing the new `react-devtools` and `react-devtools-core` packages will also require updating and coordinating with: React Native, Flipper, and Nuclide packages (and possibly others).

For now, I will be pushing my work to the [`npm-packages` branch](https://github.com/bvaughn/react-devtools-experimental/tree/npm-packages). I will rebase and cleanup before merging.

bvaughn:
Resolves #334

This PR does several things:
* Add version 4 `react-devtools` and `react-devtools-core` packages which support both React Native and e.g. Safari or iframe DOM usage.
* Replaces typed operations arrays with regular arrays in order to support Hermes. This is unfortunate, since in theory a typed array buffer could be more efficiently transferred between frontend and backend for the  web extension, but this never actually worked properly in v8, only Spidermonkey, and it fails entirely in Hermes so for the time being- it's been removed.
* Adds support for React Native (paper renderer)
* Adds a style editor for react-native and react-native-web


bvaughn:
Plowing ahead with this for now. Will follow up with some tweaks.

PR #342, opened 2019-06-30
Fix reload-and-profile attach not configurable

marvinhagemeister:
The devtools use a different way to attach the renderer with the "reload-and-profile" functionality. Before this PR it was setting it via a non-configurable object property making it impossible to set to something else. This breaks compatibility with third party clients like Preact.

Instead of going the `Object.defineProperty` way, we can just assign the property directly. Clicked a bit around in a sample CRA-App and both the standard devtools features as well as the reload-and-profile continue to work for me :tada:

### DIFF
--- a/shells/browser/shared/src/renderer.js
+++ b/shells/browser/shared/src/renderer.js
@@ -15,6 +15,9 @@ Object.defineProperty(
   '__REACT_DEVTOOLS_ATTACH__',
   ({
     enumerable: false,
+    // This property needs to be configurable to allow third-party integrations
+    // to attach their own renderer. Note that using third-party integrations
+    // is not officially supported. Use at your own risk.
     configurable: true,
     get() {
       return attach;
--- a/shells/browser/shared/src/renderer.js
+++ b/shells/browser/shared/src/renderer.js
@@ -15,6 +15,7 @@ Object.defineProperty(
   '__REACT_DEVTOOLS_ATTACH__',
   ({
     enumerable: false,
+    configurable: true,
     get() {
       return attach;
     },


shells/browser/shared/src/renderer.js
```diff
@@ -10,13 +10,4 @@
 
 import { attach } from 'src/backend/renderer';
 
-Object.defineProperty(
-  window,
-  '__REACT_DEVTOOLS_ATTACH__',
-  ({
-    enumerable: false,
```
bvaughn: The `enumerable: false` bit is important.

marvinhagemeister: Good call, I somehow missed that :+1:

marvinhagemeister:
> > Before this PR it was setting it via a non-configurable object property making it impossible to set to something else.

> 
> What is the "something else" you're referring to? Are you trying to add reload-and-profile support for Preact as well?

Yup, it's the last missing feature in our adapter for the new devtools. We still need our fair share of bug squashing, but the complete feature set is there.

For that to work we supply our own attach function that's written for Preact's internals to `__REACT_DEVTOOLS_ATTACH__`. When the reload-and-profile feature is used the extension injects its react-based renderer and prevents setting it to something else, leaving us in a broken state.

> I know it probably goes without saying, but supporting non-React renderers isn't a goal for the DevTools project :sweat_smile:

Sure thing, we're very well aware of that! To be honest we haven't run into any issues with our integration so far. For sure, we need to do a few things differently under the hood, but the API is solid and easy to integrate with and on top of that a lot easier to test against.

> It seems nice if it can be made to work without significant changes on our side. I would be open to making the property `configurable` with an inline comment about the third party use case.

Awesome, just checked and it works for our case. Updated the PR accordingly with a note like you suggested. Let me know what you think.

PR #348, opened 2019-07-13
Patch console to append component stacks

### LINKED ISSUE #347: Auto-append component stack to error logs
This will require:
* Updating React to inject `ReactDebugCurrentFrame` to DevTools (will be done as part of https://github.com/facebook/react/pull/16017 or a follow up PR)
  * Possible alternative to inject the `ReactCurrentFiber` value (https://github.com/facebook/react/pull/16133)
* Overriding `console.error`, `console.warn`, and `console.trace` to auto-append component stack (if present).

For more context see https://github.com/facebook/react/pull/16017#issuecomment-511046175

bvaughn:
- [x] Auto-append component stack to `console.error` and `console.warn` calls.
- [x] Temporarily disable all console logging before re-running a function component to inspect hooks.
- [x] Make sure not to double append a stack to any warnings logged by React.
- [x] ~~Test with multiple renderers~~ First injected renderer wins; console won't be patched for subsequent renderers

Resolves #347

Related React PR https://github.com/facebook/react/pull/16133

### Questions
* ~~Is the zero width space character (added in 6a196fd) reasonable or might it cause problems with any instrumented logging of warnings/errors?~~ This is out of DevTools hands since the component stack formatting is now being injected by React.
* Is it okay to log owner stacks rather than full component stacks? Seems preferable to me, but it _is_ different from the React official errors/warnings.

### Demo
Screen Shot 2019-07-15 at 8 52 59 AM


bvaughn:
CI is failing because of a Yarn issue, https://github.com/yarnpkg/yarn/issues/7403

bvaughn:
At this point, the tests are expected to fail until https://github.com/facebook/react/pull/16139 has landed.

src/backend/console.js
```diff
@@ -0,0 +1,85 @@
...
+          }
+        }
+      }
+
+      originalMethod(...args);
+    };
+
+    try {
+      // $FlowFixMe property error|warn is not writable.
+      targetConsole[method] = overrideMethod;
```
gaearon: One thing to be aware of is that this might cause the methods to lose location information displayed by the browser. Like when you warn/error and it says "foo.js:XX" in the upper right corner:

Screen Shot 2019-07-16 at 23 02 10

I guess it's kind of unavoidable. Although would be nice to start a conversation with browsers if we can somehow specify to skip the topmost stack level when displaying it.

(We have a similar problem in CRA already because we also override it there.)

src/hook.js
```diff
@@ -155,6 +157,27 @@ export function installHook(target: any): DevToolsHook | null {
...
+    // In that case, we'll patch later (when the frontend attaches).
+    //
+    // Don't patch in test environments because we don't want to interfere with Jest's own console overrides.
+    if (process.env.NODE_ENV !== 'test') {
+      try {
+        // We try/catch this instead of doing a typeof check,
+        // because Webpack generates JS output that will runtime error with the typeof check.
+        // e.g. typeof patchConsole === 'function'
+        // becomes: typeof _backend_console__WEBPACK_IMPORTED_MODULE_0__["patch"] === 'function'
+        // and _backend_console__WEBPACK_IMPORTED_MODULE_0__ itself is undefined
```
gaearon: Why is it undefined? I don't think I understand this.

src/backend/console.js
```diff
@@ -0,0 +1,85 @@
...
+          // $FlowFixMe We know getCurrentFiber() is a function if appendComponentStack is true.
+          let current: ?Fiber = getCurrentFiber();
+          let ownerStack: string = '';
+          while (current != null) {
+            const name = getDisplayNameForFiber(current);
+            const owner = current._debugOwner;
+            const ownerName =
+              owner != null ? getDisplayNameForFiber(owner) : null;
+
+            ownerStack += describeComponentFrame(
```
gaearon: Is there any reason we wouldn't prefer to keep this logic in React, and inject a function that gives us a full owner stack instead?

src/backend/console.js
bvaughn: Yeah, I agree with this as a downside. I will reach out to some folks on Chrome's developer tools team to at least start that conversation.

src/hook.js
bvaughn: There's a comment in the header about this too, but basically for the browser version- we inject the hook by to-stringing it.

https://github.com/bvaughn/react-devtools-experimental/blob/0f2fb5badf5908ee271e0b9e7ec7f3022e31a648/shells/browser/shared/src/injectGlobalHook.js#L85-L89

Because of this, imports won't work- although they will for React Native.

If you have suggestions on the wording I would be happy to clean it up.

src/backend/console.js
bvaughn: We've talked about this 😄Maybe you missed my reply?

https://github.com/facebook/react/pull/16139#issuecomment-511624912

src/hook.js
bvaughn: I pushed a wording improvement.

bvaughn:
I'm going to go ahead and make this a configurable preference, at least for now. It seems pretty invasive and might turn out to be annoying in ways I haven't accounted for.

toggle-component-stacks-Kapture 2019-07-17 at 9 19 17

bvaughn:
I'm going to roll forward with this, since there didn't seem to be any strong concerns (just questions). Please feel free to follow up with me if there are concerns though.

src/backend/console.js
gaearon: One way to mitigate or limit impact from this would be to have DevTools call the Global Hook when it begins or exits the React work loop. Then you could only inject the overrides at that point, and restore them afterwards.

src/backend/console.js
gaearon: Yes, what I suggested is to add this owner-only function to `ReactDebugCurrentFrame`. :-) But fair enough.

src/backend/console.js
bvaughn: Hm... do you mean the work loop would call a function on the DevTools global hook (or otherwise notify it) any time it started/stopped work?

PR #349, opened 2019-07-14
fix to input a negative number value to props

### LINKED ISSUE #346: Props editor hard to enter negative number
If you select all on a number prop and try to type in a negative value, it won't allow you.

You have to type the number in first, then add a negative sign.

koba04:
This is to fix #346.
This makes `parseFloat` delay until pressing `enter`.


bvaughn:
Thanks!

koba04:
Thanks!

PR #353, opened 2019-07-25
Feature: tweak search behaviour

### LINKED ISSUE #301: Tweak search behavior
Current search behavior is to jump to the first result in the set as you start searching. User feedback has suggested that maybe a better behavior would be to jump to the _nearest_ result (the first one after the selected index) since this would enable matches within the selected tree to be found first.

fanny:
This pr closes #301 

The strategy used here was: Get the first index where elementID is larger than the selected element and reorder the search results for these elements so that the range with the largest is at the beginning of the matrix and the range with the smallest in the end.

### DIFF
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -370,7 +370,7 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
         }
         break;
       case 'SET_SEARCH_TEXT':
-        searchIndex = 0;
+        searchIndex = null;
         searchResults = [];
         searchText = (action: ACTION_SET_SEARCH_TEXT).payload;
 
@@ -384,6 +384,8 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
               searchIndex = searchResults.findIndex(
                 value => value >= ((selectedElementID: any): number)
               );
+            } else {
+              searchIndex = 0;
             }
           }
         }
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -378,7 +378,7 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
         if (searchText !== '') {
           const regExp = createRegExp(searchText);
           store.roots.forEach(rootID => {
-            recursivelySearchTree(store, rootID, regExp, searchResults, selectedElementID, 0);
+            recursivelySearchTree(store, rootID, regExp, searchResults);
           });
 
           const firstIndexBiggerThanSelectedElement = searchResults.findIndex(
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -277,7 +277,6 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
     selectedElementIndex,
   } = state;
 
-  const prevSearchIndex = searchIndex;
   const prevSearchText = searchText;
   const numPrevSearchResults = searchResults.length;
 
@@ -383,6 +382,7 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
 
           if (searchResults.length > 0) {
             if (selectedElementID) {
+              // If there's an element selected, gets the nearest result
               searchIndex = searchResults.findIndex(
                 value => value >= selectedElementID
               );
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -784,7 +784,7 @@ function recursivelySearchTree(
   );
 }
 
-function getNearestResults (
+function getNearestResults(
   selectedElementID: number | null,
   searchResults: Array<number>
 ) {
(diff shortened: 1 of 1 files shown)


fanny:
I don't understand why your commits stay in my pr :thinking:

src/devtools/views/Components/TreeContext.js
```diff
@@ -381,13 +381,18 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
...
 
-          if (searchResults.length > 0) {
+          const searchWithNearestResults = getNearestResults(
+            selectedElementID,
+            searchResults
+          );
+
+          if (searchWithNearestResults.length > 0) {
             if (prevSearchIndex === null) {
               searchIndex = 0;
```
bvaughn: Seems like we probably don't need to create a whole new array above. I think this is the only case when we need to find the closest search match. I think we can just set a default `searchIndex` of something other than 0 if there's a `selectedElementID` and we find a "closer" match to it.

What do you think?

fanny: I don't know if I understood what you meant, can you explain it differently?

bvaughn: There's no need to create a second array. All we're doing is deciding what the best initial index is.

fanny:
I'm sorry for these many commits, I got a little lost on how I could fix an error.

bvaughn:
No worries. This looks nice. I'll try to find the time to test it soonish.

src/devtools/views/Components/TreeContext.js
```diff
@@ -380,15 +379,11 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
...
           if (searchResults.length > 0) {
-            if (prevSearchIndex === null) {
-              searchIndex = 0;
+            if (selectedElementID !== null) {
+              searchIndex = getNearestResult(searchResults, selectedElementID);
             } else {
-              searchIndex = Math.min(
-                ((prevSearchIndex: any): number),
-                searchResults.length - 1
-              );
```
bvaughn: Hm, I _think_ we might want to preserve this behavior since it would be less of a jump in some cases.

1. Go to [the demo page](https://react-devtools-experimental.now.sh/)
2. Search for "ModernContext" and hit enter twice to select the third result
3. Type an additional "." (so you're now searching for "ModernContext.")

Before this change, the selected index would go from 3 to 2 (the nearest result given the updated text):
search-before-Kapture 2019-07-27 at 8 50 10

After this change, the cursor will jump back to index 0 (the furthest result).
search-after-Kapture 2019-07-27 at 8 51 35

To be fair, this is a minor thing- but I think it's slightly nicer if we preserve the original behavior for this case. I think we can do that with just a minor tweak.

```diff
diff --git a/src/devtools/views/Components/TreeContext.js b/src/devtools/views/Components/TreeContext.js
index 5af9398..17b1a5c 100644
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -277,6 +277,7 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
     selectedElementIndex,
   } = state;
 
+  const prevSearchIndex = searchIndex;
   const prevSearchText = searchText;
   const numPrevSearchResults = searchResults.length;
 
@@ -380,10 +381,20 @@ function reduceSearchState(store: Store, state: State, action: Action): State {
             recursivelySearchTree(store, rootID, regExp, searchResults);
           });
           if (searchResults.length > 0) {
-            if (selectedElementID !== null) {
-              searchIndex = getNearestResult(searchResults, selectedElementID);
+            if (prevSearchIndex === null) {
+              if (selectedElementID !== null) {
+                searchIndex = getNearestResult(
+                  searchResults,
+                  selectedElementID
+                );
+              } else {
+                searchIndex = 0;
+              }
             } else {
-              searchIndex = 0;
+              searchIndex = Math.min(
+                ((prevSearchIndex: any): number),
+                searchResults.length - 1
+              );
             }
           }
         }
```

Since I'm already in the code I'll just go ahead and make this tweak. If you know of a good reason why we *shouldn't* do this though, let me know?

bvaughn:
Squash merged via 3a3187f

GitHub isn't picking up on the merge, so I'm going to manually close this PR. Thanks again for contributing!

bvaughn:
I think the way I merged messed up attributing the fix to you, so I force-pushed master and I'm going to merge through the GitHub UI 😄 I'll just apply my tweak after it's merged.

src/devtools/views/Components/TreeContext.js
```diff
@@ -780,4 +774,15 @@ function recursivelySearchTree(
   );
 }
 
+function getNearestResult(
+  searchResults: Array<number>,
+  selectedElementID: number | null
+) {
+  const result = searchResults.findIndex(
+    value => value >= ((selectedElementID: any): number)
```
bvaughn: Whoops! I overlooked something 😁

We shouldn't be comparing IDs here, but indices. These won't necessarily correlate. I'll make a follow up fix!

bvaughn:
Ok, index change made. Let me know if you see anything funky~

PR #354, opened 2019-07-25
[Flare] Remove the old EventComponent/EventTarget system

trueadm:
We recently redesigned the React Flare event system within React. We used to have the concepts of EventComponents and EventTargets, which have both been removed.


src/backend/renderer.js
```diff
@@ -2239,23 +2208,6 @@ export function attach(
...
-        const eventComponentInstance = node.stateNode;
-        const currentFiber = eventComponentInstance.currentFiber;
-        events.push({
-          props: eventComponentInstance.props,
-          displayName: getDisplayNameForFiber(currentFiber),
-        });
-      }
-      node = node.return;
-    }
-
```
bvaughn: I'm kind of happy to see this bit go 😁

PR #355, opened 2019-07-26
Allow Inspecting Elements Within Iframes

philipp-spiess:
This change adds support for element inspecting within `<iframe/>`s.

The iframes are kept in a Set so that we only append listeners once and clean them up properly. I’ve run a few tests and it seems to behave as expected.

Here’s a screencapture: http://cl.ly/358a8f22715d/iframe-demo.mov

The fixture includes a cross-origin iframe to make sure we do not error in this case.

I’ve run through the following scenarios locally:

- Nexted iframes
- Iframe is removed while we’re inspecting
- Chrome and Firefox

Let me know if/where I should add tests. 🙂

### DIFF
--- a/src/backend/agent.js
+++ b/src/backend/agent.js
@@ -185,13 +185,15 @@ export default class Agent extends EventEmitter<{|
 
   getIDForNode(node: Object): number | null {
     for (let rendererID in this._rendererInterfaces) {
-      // A renderer will throw if it can't find a fiber for the specified node.
-      try {
-        const renderer = ((this._rendererInterfaces[
-          (rendererID: any)
-        ]: any): RendererInterface);
-        return renderer.getFiberIDForNative(node, true);
-      } catch (e) {}
+      const renderer = ((this._rendererInterfaces[
+        (rendererID: any)
+      ]: any): RendererInterface);
+
+      const id = renderer.getFiberIDForNative(node, true);
+
+      if (id !== null) {
+        return id;
+      }
     }
     return null;
   }
--- a/src/backend/views/Highlighter/index.js
+++ b/src/backend/views/Highlighter/index.js
@@ -48,7 +48,11 @@ export default function setupHighlighter(
     hideOverlay();
     removeListenersOnWindow(window);
     iframesListeningTo.forEach(function(frame) {
-      removeListenersOnWindow(frame.contentWindow);
+      try {
+        removeListenersOnWindow(frame.contentWindow);
+      } catch (error) {
+        // This can error when the iframe is on a cross-origin.
+      }
     });
     iframesListeningTo = new Set();
   }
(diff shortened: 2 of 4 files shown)


src/backend/views/Highlighter/index.js
```diff
@@ -40,7 +46,14 @@ export default function setupHighlighter(
 
   function stopInspectingNative() {
     hideOverlay();
+    removeListenersOnWindow(window);
+    iframesListeningTo.forEach(function(frame) {
+      removeListenersOnWindow(frame.contentWindow);
```
philipp-spiess: I might want to wrap this in a try/catch as well in case the iframe becomes cross-origin after the listeners are installed.

bvaughn:
Hm... why doesn't this work on the regression test fixture? Am I overlooking something silly?

1. Run the test fixture
```sh
node ./fixtures/regression/server.js
```
2. Open the fixture: http://localhost:3000/
3. Try to select any iframe

philipp-spiess:
@bvaughn Huh this is interesting, the tooltips shows up but when you click an element, nothing happens. Perhaps a different issue given that this was also not working before but I'll look into it.

philipp-spiess:
@bvaughn Found the issue in the `getIDForNode()` implementation (see my latest comment). Inspecting in the regression fixture now works everywhere except React 0.14.9 where the element is not highlighted in DevTools. No errors in the console though so this might be ok

src/backend/agent.js
```diff
@@ -185,13 +185,15 @@ export default class Agent extends EventEmitter<{|
 
   getIDForNode(node: Object): number | null {
     for (let rendererID in this._rendererInterfaces) {
-      // A renderer will throw if it can't find a fiber for the specified node.
-      try {
-        const renderer = ((this._rendererInterfaces[
-          (rendererID: any)
-        ]: any): RendererInterface);
-        return renderer.getFiberIDForNative(node, true);
-      } catch (e) {}
```
bvaughn: Ah, nice find with the `null` check!

Although I wonder if older React versions might throw? This comment was based on this one:
https://github.com/facebook/react-devtools/blob/d839081f79f681617df7584a8ffd2f4163fa40b9/agent/Agent.js#L310-L313

Looks like error handling was added in [3979c81](https://github.com/facebook/react-devtools/commit/3979c811d5cf19da9924a9aff33fb1484adb6a0d#diff-7d7518fc7bed10b7a64646d6b363b99aR225) and the comment was later added in [11537a0](https://github.com/facebook/react-devtools/commit/11537a04ffda71b0c75ca5346354d0df49a3bc89#diff-7d7518fc7bed10b7a64646d6b363b99aR269). While DevTools itself doesn't throw, maybe e.g. `nativeTagToRootNodeID` did for some versions?

Might be best to re-add the `try`/`catch` too, just in case. I can do that after merging!

philipp-spiess: Makes sense, thanks! 🙂

PR #360, opened 2019-08-01
feat: match HOC display names during search

### LINKED ISSUE #299: Support search by HOC name
PR #295 removed the HOC portion of the name from element `displayName`. As a result, you can no longer search by HOC name. We could re-add this by updating the search logic to also check the `hocDisplayNames` strings.

SavePointSam:
closes #299

Enables the user to use the component search capability against HOC display names.

### DIFF
--- a/src/devtools/views/Components/TreeContext.js
+++ b/src/devtools/views/Components/TreeContext.js
@@ -773,14 +773,20 @@ function recursivelySearchTree(
   regExp: RegExp,
   searchResults: Array<number>
 ): void {
-  const { children, displayName } = ((store.getElementByID(
+  const { children, displayName, hocDisplayNames } = ((store.getElementByID(
     elementID
   ): any): Element);
-  if (displayName !== null) {
-    if (regExp.test(displayName)) {
-      searchResults.push(elementID);
-    }
+
+  if (displayName != null && regExp.test(displayName) === true) {
+    searchResults.push(elementID);
+  } else if (
+    hocDisplayNames != null &&
+    hocDisplayNames.length > 0 &&
+    hocDisplayNames.some(name => regExp.test(name)) === true
+  ) {
+    searchResults.push(elementID);
   }
+
   children.forEach(childID =>
     recursivelySearchTree(store, childID, regExp, searchResults)
   );
--- a/src/__tests__/__snapshots__/treeContext-test.js.snap
+++ b/src/__tests__/__snapshots__/treeContext-test.js.snap
@@ -558,12 +558,13 @@ exports[`TreeListContext search state should find elements matching search text:
     <Foo>
     <Bar>
     <Baz>
+    <Qux> [withHOC]
 `;
 
 exports[`TreeListContext search state should find elements matching search text: 1: initial state 1`] = `
 Object {
   "inspectedElementID": null,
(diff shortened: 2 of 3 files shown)


src/devtools/views/Components/TreeContext.js
```diff
@@ -773,14 +773,34 @@ function recursivelySearchTree(
...
+    pushResult();
+  }
+
+  if (
+    resultWasPushed === false &&
+    hocDisplayNames != null &&
+    hocDisplayNames.length > 0
+  ) {
+    if (hocDisplayNames.some(name => regExp.test(name) === true) === true) {
+      pushResult();
```
bvaughn: Seems like we could simplify this a bit and remove the overhead of an extra function call by using an if/else. What do you think?

```js
function recursivelySearchTree(
  store: Store,
  elementID: number,
  regExp: RegExp,
  searchResults: Array<number>
): void {
  const { children, displayName, hocDisplayNames } = ((store.getElementByID(
    elementID
  ): any): Element);

  if (
    displayName != null &&
    regExp.test(displayName) === true
  ) {
    searchResults.push(elementID);
  } else if (
    hocDisplayNames != null &&
    hocDisplayNames.length > 0 &&
    hocDisplayNames.some(name => regExp.test(name))
  ) {
    searchResults.push(elementID);
  }

  children.forEach(childID =>
    recursivelySearchTree(store, childID, regExp, searchResults)
  );
}
```

SavePointSam: Great!

SavePointSam:
> This feature seems like it would benefit from a couple of new tests 😄

I was wanting to add tests anyway. I’ll get on them. :D

> After trying the feature, I also think that maybe some form of visual results highlighting is going to be important after all, to avoid confusion

Yeah, I felt the same about this. It’s certainly not clear why an element is being returned as a match.

> Maybe something like below?

I like this a lot! Solves the problem I was hitting with trying to show the user the match is from a tag that was being truncated. I’d be happy to look into adding this as well.

SavePointSam:
I've simplified the match logic and added/updated related tests. I can add the highlighting tonight or I can add it as a follow-up PR if you like.

bvaughn:
One thing I notice now is that maybe we should special case HOCs like forward ref and memo so the search matches them too, because it seems a little weird that you can search for e.g. "withFoo" but not "Memo" when they look the same.

SavePointSam:
Sounds good!

bvaughn:
Thanks for contributing by the way! 😁

SavePointSam:
> it's going to need a little thought on how to most efficiently determine which part of the element (name, visible HOC name, or +number) to highlight

I would image there is just a preference system where the preferred match is the one that is surfaced. I would also suggest that we match on all so that the behavior of the system is very explicit, however that approach is _slightly_ misleading as the actual match algorithm is preference based for quick matching. This is probably worth a ticket you can assign to me, if you like.

> you can search for e.g. "withFoo" but not "Memo" when they look the same

Oh! This is a very good idea, probably worth a ticket if you want to assign it to me.

> Thanks for contributing by the way!

My pleasure! I'm a big fan of this project. :D

bvaughn:
Follow up work briefly detailed in #365

PR #364, opened 2019-08-03
Refactor inspect/select logic so that $r contains hooks data

### LINKED ISSUE #357: How to get hook state in console?
I am trying to interactively examine data I'm receiving from an apollo query via a hook but I cannot access it via $r.

image

With an HOC I'd be able to directly see the prop in the console, write filters against it to track down needles in haystacks as I'm debugging, etc.  

A way to get to this from the console would be greatly appreciated.  Additionally, it would be nice to be able to somehow get a reference for the hook under my mouse - something like $r.hooks.thing.deep.path.

I am absolutely **loving** hooks in react and have been using it for all new development within my company (top Fashion PLM software company).  React is far and away the best ecosystem I have ever encountered for general application development.

bvaughn:
Closes #357

This change enables a few things:
* Removes a redundant `Bridge` message.
* Enables `$r` to include hooks info for function components.
* Ensures `$r` is periodically updated if the selected element is re-rendered.

One thing to note is that the `hooks` value exposed is the internal data structure created by [`react-debug-hooks`](https://github.com/bvaughn/react-devtools-experimental/blob/master/src/backend/ReactDebugHooks.js) so it isn't necessarily as parsable or useful as the `props` or `state` objects. Still seems better than nothing though?

Screen Shot 2019-08-03 at 5 55 02 PM

### DIFF
--- a/packages/react-devtools-core/src/backend.js
+++ b/packages/react-devtools-core/src/backend.js
@@ -112,8 +112,8 @@ export function connectToDevTools(options: ?ConnectOptions) {
       },
     });
     bridge.addListener(
-      'selectElement',
-      ({ id, rendererID }: {| id: number, rendererID: number |}) => {
+      'inspectElement',
+      ({ id, rendererID }: { id: number, rendererID: number }) => {
         const renderer = agent.rendererInterfaces[rendererID];
         if (renderer != null) {
           // Send event for RN to highlight.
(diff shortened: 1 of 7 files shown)


bvaughn:
Eh, I think this is probably a reasonable change. Rolling forward with it.

PR #376, opened 2019-08-12
Add data-ids for tutorial automation

bvaughn:
Not sure how I feel about this yet, if it's worth the added complexity and coupling.

https://github.com/bvaughn/react-devtools-tutorial/pull/1


bvaughn:
I don't think this is worth the added complexity and coupling. Going to close for now.

PR #385, opened 2019-08-24
Refactored backend to remove recursion

bvaughn:
DevTools doesn't currently handle extremely deep trees (https://github.com/facebook/react/issues/16491) or extremely wide trees (https://github.com/facebook/react/issues/16501) very well, due to recursion in the backend interface. This PR removes _most_ of that.

As a sanity test, I confirmed that after this refactor, DevTools was able to handle the following:
```js
const Child = ({ children = null }) => children;

const Deep = () => {
  let children = null;
  for (let i = 0; i < 15000; i++) {
    children = <Child>{children}</Child>;
  }
  return children;
};

const Wide = () => {
  let children = [];
  for (let i = 0; i < 15000; i++) {
    children.push(<Child key={i} />);
  }
  return children;
};
```

### DIFF
--- a/src/backend/renderer.js
+++ b/src/backend/renderer.js
@@ -1213,6 +1213,9 @@ export function attach(
     }
   }
 
+  // TRICKY
+  // This method recursively mounts a fiber, but it does so iteratively.
+  // This is to avoid hitting the call stack limit for extremely deep or wide trees.
   function mountFiberRecursively(
     fiber: Fiber,
     parentFiber: Fiber | null,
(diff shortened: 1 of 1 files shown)


src/backend/renderer.js
```diff
@@ -1399,22 +1439,37 @@ export function attach(
...
+        }
       }
     }
   }
 
   // Returns whether closest unfiltered fiber parent needs to reset its child list.
+  //
+  // TRICKY
+  // Although this method is recursive, refactoring it to be iterative would add a lot of complexity.
+  // Since it is not expected to operate on huge parts of the tree at once, it's probably okay?
```
gaearon, line 1472: Why is it not expected?

bvaughn: Maybe I misread the function? 🙂 I'm not as familiar with the suspense logic since I didn't write it. I just skimmed it and read the header comment:
> // Returns whether closest unfiltered fiber parent needs to reset its child list.

It seemed like it didn't recurse as greedily as the mount/unmount functions. Looking at our tests it also seemed like the function didn't recurse as deeply as the others (primarily mount) but that could just be due to it being a small sample size.

Think we should rewrite it also? I would have just done it initially, except that it seemed trickier than the other two since it also depends on its own return value.

kaykayehnn: It seems that traversing deep trees causes this function to stack overflow. See this example

```jsx
const Child = ({ children = null }) => children;

const Deep = () => {
  const [condition, setCondition] = useState(false);

  let children = null;
  for (let i = 0; i < 15000; i++) {
    children = <Child>{children}</Child>;
  }
  return (
    <div>
      <button onClick={() => setCondition(c => !c)}>Rerender</button>
      {children}
    </div>
  );
};
```

The initial render works, but rerendering the component recurses too deeply and overflows. Adding the same rerender button on the `Wide` component works fine though.

bvaughn:
Going to migrate this PR to the React repo -> https://github.com/facebook/react/pull/16627

